PR #8, opened 2021-06-16
[iOS Object Detection] Add object detection e2e ios sample application

YUNQIUGUO:
- A basic application works as an object detector using `ssd_mobilenet_v1(quantized)` model (from TF)

Several todos: ✅
1. Complete the readme to add documentations for the part `convert model from TF->ONNX->ORT format` 
2. Add host link for the prepared ort ssd mobilenet model for user to download 
3. Quantized model is not supported for this sample app


mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+            return nil
+        }
+        
+        // Start the ORT inference environment
+        let env = try ORTEnv(loggingLevel: ORTLoggingLevel.warning)
+        let options = try ORTSessionOptions()
+        try options.setLogSeverityLevel(ORTLoggingLevel.verbose)
+        try options.setIntraOpNumThreads(self.threadCount) // TODO: check if calling the right methods
+        
+        let session = try ORTSession(env: env, modelPath: modelPath, sessionOptions: options)
```
gwang-msft, line 118: Move these into init() and make session a class variable?
Such that you don't need to create a session for every frame?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
edgchen1, line 4: why both MIT and Apache licenses?

YUNQIUGUO: Because some of the code are from files licensed under Apache licenses. You suggest only add Apache licenses?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+        case kCVPixelFormatType_32ARGB:
+            vImageConvert_ARGB8888toRGB888(&sourceBuffer, &destinationBuffer, UInt32(kvImageNoFlags))
+        case kCVPixelFormatType_32RGBA:
+            vImageConvert_RGBA8888toRGB888(&sourceBuffer, &destinationBuffer, UInt32(kvImageNoFlags))
+        default:
+            // Unknown pixel format.
+            return nil
+        }
+        
+        let byteData = Data(bytes: destinationBuffer.data, count: destinationBuffer.rowBytes * height)
```
edgchen1, line 395: looks like there are some copies, e.g. destinationBuffer -> byteData and floats -> return value. is it possible to remove some of these?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+                                      runOptions: try ORTRunOptions())
+        interval = Date().timeIntervalSince(startDate) * 1000
+        
+        guard let rawOutputValue = outputs["TFLite_Detection_PostProcess"] else {
+            throw OrtModelError.error("failed to get model output_0")
+        }
+        
+        let rawOutputData = try rawOutputValue.tensorData() as Data
+        
+        guard let outputArr: [Float32] = arrayCopiedFromData(rawOutputData) else {
```
edgchen1, line 151: is it possible to avoid copying the output data? maybe it doesn't matter so much if we know the data size is small

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+                                       elementType: ORTTensorElementDataType.float,
+                                       shape: [1, 300, 300, 3])
+        // Run ORT InferenceSession
+        let startDate = Date()
+        let outputs = try session.run(withInputs: [inputName: inputTensor],
+                                      outputNames: ["TFLite_Detection_PostProcess",
+                                                    "TFLite_Detection_PostProcess:1",
+                                                    "TFLite_Detection_PostProcess:2",
+                                                    "TFLite_Detection_PostProcess:3"],
+                                      runOptions: try ORTRunOptions())
```
edgchen1, line 142: could also pass runOptions: nil

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+        }
+        var scaledVImageBuffer = vImage_Buffer(data: scaledImageBytes, height: UInt(size.height), width: UInt(size.width), rowBytes: scaledRowBytes)
+        
+        // Perform the scale operation on input image buffer and store it in scaled vImage buffer.
+        let scaleError = vImageScale_ARGB8888(&inputVImageBuffer, &scaledVImageBuffer, nil, vImage_Flags(0))
+        
+        CVPixelBufferUnlockBaseAddress(buffer, CVPixelBufferLockFlags(rawValue: 0))
+        
+        guard scaleError == kvImageNoError else {
+            return nil
```
edgchen1, line 281: could scaledImageBytes get leaked here?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+        ofSize size: CGSize,
+        _ buffer: CVPixelBuffer
+    ) -> CVPixelBuffer? {
+        
+        let imageWidth = CVPixelBufferGetWidth(buffer)
+        let imageHeight = CVPixelBufferGetHeight(buffer)
+        let pixelBufferType = CVPixelBufferGetPixelFormatType(buffer)
+        
+        assert(pixelBufferType == kCVPixelFormatType_32BGRA ||
+                pixelBufferType == kCVPixelFormatType_32ARGB)
```
edgchen1, line 252: nit: formatting alignment
I tried swift-format and it seems to work pretty well
https://github.com/apple/swift-format

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+    
+    init?(threadCount: Int32 = 1) {
+
+        self.threadCount = threadCount
+        
+        super.init()
+    }
+
+    // This methods preprocess the image, runs the ort inferencesession and returns the inference result
+    func runModel(onFrame pixelBuffer: CVPixelBuffer, modelFileInfo: FileInfo, labelsFileInfo: FileInfo) throws -> Result? {
```
edgchen1, line 81: general comment: I noticed there are a few different error reporting modes - e.g., returning nil/printing, throwing, calling fatalError(). I'm not too familiar with Swift conventions. is there a reason for using each of the different modes?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+        UIColor.purple,
+        UIColor.magenta,
+        UIColor.yellow,
+        UIColor.cyan,
+        UIColor.brown
+    ]
+    
+    private var labels: [String] = []
+    
+    init?(threadCount: Int32 = 1) {
```
edgchen1, line 73: why `init?` instead of `init`?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+            scaledPixelBuffer,
+            byteCount: batchSize * inputWidth * inputHeight * inputChannels
+        ) else {
+            print("Failed to convert the image buffer to RGB data.")
+            return nil
+        }
+        
+        let inputTensor = try ORTValue(tensorData: NSMutableData(data: rgbData),
+                                       elementType: ORTTensorElementDataType.float,
+                                       shape: [1, 300, 300, 3])
```
edgchen1, line 134: use batchSize, inputWidth... here?

mobile/examples/object_detections/iOS/README.md
```diff
@@ -0,0 +1,45 @@
+# ONNX Runtime Mobile object detection iOS sample application
+
+This is an example app uses object detection which is able to continuously detect the objects in the frames seen by your iOS device's back camera and display the detected object bounding boxes, detected class and corresponding inference confidence on the screen.
```
edgchen1, line 3: "This is an example app **uses** object detection"
one possible change: "This is an example app using object detection"

mobile/examples/object_detections/iOS/README.md
```diff
@@ -0,0 +1,45 @@
...
+- Clone the `onnxruntime-inference-examples` source code repo
+
+## Build And Run
+
+0. [Optional] Prepare ORT format model from TF -> ONNX -> ORT
+[//]: # (Add the information from TF->ONNX->ORT here)
+
+1. Install CocoaPods. `sudo gem install cocoapods`
+
+2. Run `pod install` to generate the workspace file under `examples/object_detections/ios/`.
```
edgchen1, line 25: does this assume the current directory is `examples/object_detections/ios/`? might be good to state that if so

mobile/examples/object_detections/iOS/README.md
```diff
@@ -0,0 +1,45 @@
...
+5. Connect your iOS device, build and run the app. You'll have to grant permissions for the app to use the device's camera.
+
+
+### iOS App related information
+
+This app uses [ONNX Runtime Objective-C API](https://www.onnxruntime.ai/docs/reference/api/objectivec-api.html) for performing object detection function.
+
+It is written entirely in Swift and uses a bridgingheader file for the framework to be used in a Swift app.
+
+#
```
edgchen1, line 42: extra `#`?

mobile/examples/object_detections/iOS/README.md
```diff
@@ -0,0 +1,45 @@
...
+Here's an example screenshot of the app:
+
+<img width=20% src="images/objdetect_screenshot.jpg" alt="App Screenshot" />
```
edgchen1, line 45: just noticed that in the screenshot the "mouse" box seems to be above the actual mouse

mobile/examples/object_detections/iOS/README.md
```diff
@@ -0,0 +1,45 @@
+# ONNX Runtime Mobile object detection iOS sample application
```
edgchen1, line 1: how about renaming the directory `mobile/examples/object_detections` to `mobile/examples/object_detection`?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
edgchen1: oh, I see - was it copied from some other project and modified?
we've also done something like this:
https://github.com/microsoft/onnxruntime/blob/1d8edd0b5bcf3c3172cc10ae60662073e8f8f748/include/onnxruntime/core/common/common.h#L1-L16
I'm not sure what the proper way to handle it is though

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,463 @@
...
+        var inputVImageBuffer = vImage_Buffer(
+            data: inputBaseAddress, height: UInt(imageHeight), width: UInt(imageWidth),
+            rowBytes: inputImageRowBytes)
+        
+        let scaledRowBytes = Int(size.width) * imageChannels
+        guard  let scaledImageBytes = malloc(Int(size.height) * scaledRowBytes) else {
+            return nil
+        }
+        var scaledVImageBuffer = vImage_Buffer(data: scaledImageBytes, height: UInt(size.height), width: UInt(size.width), rowBytes: scaledRowBytes)
+
```
gwang-msft, line 274: Use vImageBuffer_Init?

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
YUNQIUGUO: I am not sure there really exists an error handling convention in Swift or not. But in general I think 
- fatalError() for my case is only used for failures like AVCameraSession failure/videodeviceinputoutput/resourcefile fetch failure, which can cause a crash for the app. Part of this is directly kept from apple AVCamera Sample app.
- for printing/return nil, it is used in pair with `guard let`, which is for some variable type match, value checking, verify conditions, etc. Or with do/catch handling.
-  for throw/throws, I believe it is mainly used just for throwing OrtModelError in the app.

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
YUNQIUGUO: Originally I thought I can create a failable initializer for modelhandler, so that I can put all the loading models and label files (which can `return nil` therefore requires a failable init) and OrtSession initialization in it. But tried for several times it causes the app hung from execution. Still looking into it.

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
edgchen1: I guess what I'm asking is - could it be more consistent or is there a good reason to do all three?

YUNQIUGUO: oh okay, sorry about the misunderstanding. Maybe I can try to only use the previous two to make it more consistent and reduces the number of printing messages that can cause some confusion.

mobile/README.md
```diff
@@ -13,3 +13,9 @@ The example app shows basic usage of the ORT APIs.
 The example app uses image classification which is able to continuously classify the objects it sees from the device's camera in real-time and displays the most probable inference results on the screen.
 
 - [Android Image Classifier](examples/image_classifications/android)
+
+## Object Detection
+
+The example app uses object detection which is able to continuously detect the objects in the frames seen by your iOS device's back camera and display the detected object bounding boxes, detected class and corresponding inference confidence on the screen.
+
+- [iOS Object Detector](examples/object_detections/ios)
```
gwang-msft, line 21: (examples/object_detection/ios)

mobile/examples/object_detection/ios/README.md
```diff
@@ -0,0 +1,48 @@
...
+- Xcode command line tools `xcode-select --install`
+- Clone the `onnxruntime-inference-examples` source code repo
+
+## Build And Run
+
+0. [Optional] Prepare ORT format model from TF -> ONNX -> ORT
+    - Open [Converting SSD Mobilenet from Tensorflow to ONNX](https://github.com/onnx/tensorflow-onnx/blob/master/tutorials/ConvertingSSDMobilenetToONNX.ipynb). This notebook will demostrate you how to convert pre-trained ssd models from TF to onnx.
+    - Once you get the model in onnx format, e.g. `ssd_mobilenet_v2_float.onnx`, you can follow [Convert ONNX models to ORT format](https://www.onnxruntime.ai/docs/how-to/mobile/model-conversion.html#converting-onnx-models-to-ort-format) to get an ORT format model.
+
+    Note: This step is optional. A converted ort format model is provided [here](https://microsoft-my.sharepoint.com/personal/guorachel_microsoft_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fguorachel%5Fmicrosoft%5Fcom%2FDocuments%2FShared%2FPublic) along with an onnx model format file.
```
gwang-msft, line 24: Maybe try the direct link without personal information

edgchen1: should we consider putting it in one of the Azure storage accounts? not sure if we already have a place for example models, but maybe it would be good to have one.

YUNQIUGUO: Yes. It would be better to put it in a more official host link.

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,448 @@
...
+
+    // This method postprocesses the results including processing bounding boxes, sort detected scores, etc. and returns a result array
+    func formatResults(detectionBoxes: [Float32], detectionClasses: [Float32], detectionScores: [Float32], numDetections: Int, width: CGFloat, height: CGFloat) -> [Inference] {
+        var resultsArray: [Inference] = []
+        
+        if numDetections == 0 {
+            return resultsArray
+        }
+        
+        for i in 0 ... numDetections - 1 {
```
edgchen1, line 200: >0 ... numDetections - 1 [](http://example.com/codeflow?start=17&length=23)

could be `0..<numDetections`

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,448 @@
...
+        }
+        
+        // Get vImage_buffer
+        var inputVImageBuffer = vImage_Buffer(
+            data: inputBaseAddress, height: UInt(imageHeight), width: UInt(imageWidth),
+            rowBytes: inputImageRowBytes
+        )
+        
+        let scaledRowBytes = Int(size.width) * imageChannels
+        guard let scaledImageBytes = malloc(Int(size.height) * scaledRowBytes) else {
```
edgchen1, line 266: >malloc [](http://example.com/codeflow?start=37&length=6)

one idea to easily handle the free - how about having `defer { free(scaledImageBytes) }` and setting scaledImageBytes to nil after handing it over to CVPixelBufferCreateWithBytes()? free(0) (and, I assume, free(nil)) should be a no-op

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,448 @@
...
+            return nil
+        }
+        var scaledVImageBuffer = vImage_Buffer(data: scaledImageBytes, height: UInt(size.height), width: UInt(size.width), rowBytes: scaledRowBytes)
+        
+        // Perform the scale operation on input image buffer and store it in scaled vImage buffer.
+        let scaleError = vImageScale_ARGB8888(&inputVImageBuffer, &scaledVImageBuffer, nil, vImage_Flags(0))
+        
+        CVPixelBufferUnlockBaseAddress(buffer, CVPixelBufferLockFlags(rawValue: 0))
```
edgchen1, line 274: >CVPixelBufferUnlockBaseAddress [](http://example.com/codeflow?start=8&length=30)

should we call this from a defer block?
as a side note, I found we can introduce a new scope with `do { ... }` to run the defer block sooner if needed

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
YUNQIUGUO: Not so sure how to do the part "set the scaledImageBytes to nil.." because I guess I set scaledImageBytes to be a `let`. But thanks for the suggestion,I understand that we should call free wherever a possible error can cause exit. I added defer{free(scaledImageBytes)} inside`guard scaleError == kvImageNoError`.

edgchen1: right - scaledImageBytes would have to be declared with var instead

mobile/examples/object_detection/ios/ORTObjectDetection/download.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# ============================================================
+
+MODELS_URL="https://tfhub.dev/tensorflow/lite-model/ssd_mobilenet_v1/1/metadata/1?lite-format=tflite"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+cd "$SCRIPT_DIR"
```
edgchen1, line 9: is it necessary to `cd`?

YUNQIUGUO: Right. This can be omitted. Also will delete the `ModelsAndData` dir later as the labels data and model can all be generated by the script.

mobile/examples/object_detection/ios/ORTObjectDetection/download.sh
```diff
@@ -0,0 +1,26 @@
...
+
+mkdir -p "${SCRIPT_DIR}/ModelsAndData"
+
+DOWNLOAD_DIR="${SCRIPT_DIR}/ModelsAndData"
```
edgchen1, line 13: nit: could put this first and then call `mkdir -p "${DOWNLOAD_DIR}"`

mobile/examples/object_detection/ios/ORTObjectDetection/download.sh
```diff
@@ -0,0 +1,26 @@
...
+
+#Download source tflite model file
+curl -L ${MODELS_URL} >${DOWNLOAD_DIR}/ssd_mobilenet_v1_1_metadata_1.tflite
+
+#Unzip and get model metadata
+curl -L ${MODELS_URL} >${DOWNLOAD_DIR}/ssd_mobilenet_v1_1_metadata_1.zip
```
edgchen1, line 19: could we just download it once and unzip ssd_mobilenet_v1_1_metadata_1.tflite?

mobile/examples/object_detection/ios/README.md
```diff
@@ -0,0 +1,49 @@
...
+- Clone the `onnxruntime-inference-examples` source code repo
+
+## Build And Run
+
+1. Install CocoaPods. `sudo gem install cocoapods`
+
+2. Run `pod install` to generate the workspace file under `<ONNXRuntime-inference-example-root>/mobile/examples/object_detections/ios/`. 
+- At the end of this step, you should get a file called `ORTObjectDetection.xcworkspace`.
+
+3. Run download script `download.sh` under `<ONNXRuntime-inference-example-root>/mobile/examples/object_detections/ios/ORTObjectDetection/`. The script will download an original tf model along with the model metadata `labelmap.txt` and convert the model to onnx and then further convert it to ort format model (the format can be executed on mobile applications).
```
skottmckay, line 26: nit: as a naive user I need a bit more info here, like where do I run each command from. is it from a terminal window or within xcode? if a terminal, do commands like `pod install` need to run from a specific directory?

mobile/examples/object_detection/ios/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### iOS App related information
+
+This app uses [ONNX Runtime Objective-C API](https://www.onnxruntime.ai/docs/reference/api/objectivec-api.html) for performing object detection functionality.
+
+It is written entirely in Swift and uses a bridgingheader file for the framework to be used in a Swift app.
+
+### Related Resources
+- [Converting SSD Mobilenet from Tensorflow to ONNX](https://github.com/onnx/tensorflow-onnx/blob/master/tutorials/ConvertingSSDMobilenetToONNX.ipynb). 
+
```
skottmckay, line 43: I probably wouldn't link to this as I think it potentially adds more confusion than it's worth. A link to the tf2onnx documentation on converting tflite models would however be helpful.

mobile/examples/object_detection/ios/ORTObjectDetection/download.sh
```diff
@@ -0,0 +1,22 @@
...
+curl -L ${MODELS_URL} >${DOWNLOAD_DIR}/ssd_mobilenet_v1_1_metadata_1.tflite
+
+#Unzip and get model metadata
+unzip ${DOWNLOAD_DIR}/ssd_mobilenet_v1_1_metadata_1.tflite -d ${DOWNLOAD_DIR}
+
+#Convert tflite model to onnx model
+python3 -m tf2onnx.convert --tflite ${DOWNLOAD_DIR}/ssd_mobilenet_v1_1_metadata_1.tflite --opset 13 --output ${DOWNLOAD_DIR}/ssd_mobilenet_v1.onnx
+
+#Convert onnx model to ort format
+python3 -m onnxruntime.tools.convert_onnx_models_to_ort ${DOWNLOAD_DIR}
```
skottmckay, line 22: Would be good to mention in the comment what the output of this command is (the .ort file and its location) so the user has a hint of where to look and what they're looking for.

mobile/examples/object_detection/ios/ORTObjectDetection/CameraManager.swift
```diff
@@ -0,0 +1,212 @@
...
+import UIKit
+
+// MARK: CameraFeedManagerDelegate Declaration
+
+protocol CameraManagerDelegate: AnyObject {
+    // This method delivers the pixel buffer of the current frame seen by the device's camera.
+    func didOutput(pixelBuffer: CVPixelBuffer)
+    
+    // This method initimates that the camera permissions have been denied.
+    func presentCameraPermissionsDeniedAlert()
```
skottmckay, line 28: not sure what 'initimates' means

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
```diff
@@ -0,0 +1,442 @@
...
+        
+        // Convert the scaled vImage buffer to CVPixelBuffer
+        let conversionStatus = CVPixelBufferCreateWithBytes(
+            nil, Int(size.width), Int(size.height), pixelBufferType, scaledImageBytes,
+            scaledRowBytes, releaseCallBack, nil, nil, &scaledPixelBuffer
+        )
+        
+        guard conversionStatus == kCVReturnSuccess else {
+            free(scaledImageBytes)
+            return nil
```
skottmckay, line 300: nit: is there a way to avoid the multiple calls to `free(scaledImageBytes)`?

mobile/examples/object_detection/ios/ORTObjectDetection/OverlayView.swift
```diff
@@ -0,0 +1,69 @@
...
+        let stringBgRect = CGRect(x: objectOverlay.borderRect.origin.x, y: objectOverlay.borderRect.origin.y, width: 2 * stringHorizontalSpacing + objectOverlay.nameStringSize.width, height: 2 * stringVerticalSpacing + objectOverlay.nameStringSize.height)
+        
+        let stringBgPath = UIBezierPath(rect: stringBgRect)
+        objectOverlay.color.withAlphaComponent(stringBgAlpha).setFill()
+        stringBgPath.fill()
+    }
+    
+    func drawName(of objectOverlay: ObjectOverlay) {
+        let stringRect = CGRect(x: objectOverlay.borderRect.origin.x + stringHorizontalSpacing, y: objectOverlay.borderRect.origin.y + stringVerticalSpacing, width: objectOverlay.nameStringSize.width, height: objectOverlay.nameStringSize.height)
+
```
skottmckay, line 65: nit: can we keep the line lengths under 120 like the rest of our code? something like this may be helpful. don't care if it's not automated though. [https://realm.github.io/SwiftLint/line_length.html](https://realm.github.io/SwiftLint/line_length.html)

mobile/examples/object_detection/ios/ORTObjectDetection/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -0,0 +1,106 @@
+{
+  "images" : [
+    {
```
skottmckay, line 3: :heart: the icons

mobile/examples/object_detection/ios/ORTObjectDetection/download.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
gwang-msft, line 1: Maybe this file can be renamed to prepare_model.sh?
Also do you want to delete the tflite model after the coversion?

mobile/examples/object_detection/ios/Podfile.lock
```diff
@@ -0,0 +1,22 @@
+PODS:
```
gwang-msft, line 1: Add this to the .gitignore?

mobile/examples/object_detection/ios/README.md
```diff
@@ -0,0 +1,49 @@
...
+1. Install CocoaPods. `sudo gem install cocoapods`
+
+2. Run `pod install` to generate the workspace file under `<ONNXRuntime-inference-example-root>/mobile/examples/object_detections/ios/`. 
+- At the end of this step, you should get a file called `ORTObjectDetection.xcworkspace`.
+
+3. Run download script `download.sh` under `<ONNXRuntime-inference-example-root>/mobile/examples/object_detections/ios/ORTObjectDetection/`. The script will download an original tf model along with the model metadata `labelmap.txt` and convert the model to onnx and then further convert it to ort format model (the format can be executed on mobile applications).
+- At the end of this step, you should get a directory `ModelsAndData` which contains the ort format model `ssd_mobilenet_v1.all.ort` and model label data file `labelmap.txt`.
+
+    Note: The model generated might need to be copied to app bundle. i.e. In Xcode, `Build phases-Expand Copy Bundle Resources-Click '+' and select model file name ssd_mobilenet_v1.all.ort`.
```
gwang-msft, line 29: Do you need to copy the labelmap.txt as well?

mobile/examples/object_detection/ios/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
+
+source 'https://github.com/CocoaPods/Specs.git'
```
gwang-msft, line 4: Do we need this?

mobile/examples/object_detection/ios/ORTObjectDetection/OverlayView.swift
edgchen1: FYI, we have a config file for [swift-format](https://github.com/apple/swift-format)

mobile/examples/object_detection/ios/Podfile
YUNQIUGUO: yes, this is necessary.:D

mobile/examples/object_detection/ios/ORTObjectDetection/Assets.xcassets/AppIcon.appiconset/Contents.json
gwang-msft: All of our example apps should use this

mobile/examples/object_detection/ios/Podfile.lock
YUNQIUGUO: oh I saw in the tensorflow examples, all of their `podfile.lock` is left there. so figured it might be okay.

mobile/examples/object_detection/ios/ORTObjectDetection/ModelHandler.swift
YUNQIUGUO: unfortunately, it seems like we have to manually call `free(scaledImageBytes)` at every exits..

mobile/examples/object_detection/ios/Podfile
```diff
@@ -0,0 +1,11 @@
...
+
+target 'ORTObjectDetection' do
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
+  # Pods for ORTObjectDetection
+  pod 'onnxruntime-mobile-objc', '1.8.1'
```
edgchen1, line 10: >, '1.8.1' [](http://example.com/codeflow?start=31&length=9)

now that we have a release version, should we always use the latest (i.e., not specify a version here)? maybe we can do that after setting up a CI test for this example, that can be done later

mobile/examples/object_detection/ios/Podfile
edgchen1: hm, I managed to have a Podfile without it. could be that it took awhile for the pod to go to the CDN after it was initially released and this was briefly necessary

YUNQIUGUO: oh okay I see. will update later then.

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
edgchen1: I think this is a good idea. Looks like maybe this comment was resolved before it got addressed?

skottmckay: More than a good idea - we don't want people copying this and creating a new session for each run as that's hugely expensive and will result in horrible performance.

gwang-msft (approved):

mobile/examples/object_detections/iOS/ORTObjectDetection/ModelHandler.swift
YUNQIUGUO: Sorry the comments just got missed. Also noticed couple typos. Will update together in another pr.

YUNQIUGUO: After moving the session to initialization- init(), the `session.run()`will fail. Spent some time looking at it but couldn't figure out the reason. The ORTSession and modelhandler initializations are successul. Error occurs whenever we call `session.run(withInputs, outputNames, runoptions...)`...

The changes:
[https://github.com/microsoft/onnxruntime-inference-examples/compare/yguo/fix-ios-objdetect](https://github.com/microsoft/onnxruntime-inference-examples/compare/yguo/fix-ios-objdetect)

Appreciate any suggestions on potential reasons that could cause this error

gwang-msft: Try move the env to class variable also

YUNQIUGUO: > Try move the env to class variable also

Thanks!! the app works and I updated.

PR #9, opened 2021-06-22
[iOS] Add SpeechRecognition example

edgchen1:
Add an iOS speech recognition example that uses the Wav2Vec 2.0 model.


mobile/examples/speech_recognition/ios/SpeechRecognition/AudioRecorder.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    init(callback: @escaping RecordingDoneCallback) {
+      self.callback = callback
+    }
+
+    func audioRecorderDidFinishRecording(
+      _ recorder: AVAudioRecorder,
+      successfully flag: Bool
+    ) {
+      print("audioRecorderDidFinishRecording")
```
YUNQIUGUO, line 76: Is this a message will be displayed on the screen? or just used for debugging the app or so..

mobile/examples/speech_recognition/ios/readme.md
```diff
@@ -0,0 +1,35 @@
...
+
+From this directory, run:
+
+```bash
+pod install
+```
+
+## Build and run
+
+Open the generated SpeechRecognition.xcworkspace file in Xcode to build and run the example.
```
YUNQIUGUO, line 35: >  Xcode 
nit: maybe we can add information about xcode/ios version in the prerequisite/requirements section?

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
```diff
@@ -0,0 +1,91 @@
...
+      elementType: ORTTensorElementDataType.float,
+      shape: inputShape)
+
+    let startTime = DispatchTime.now()
+    let outputs = try ortSession.run(
+      withInputs: ["input": input],
+      outputNames: ["output"],
+      runOptions: nil)
+    let endTime = DispatchTime.now()
+    print("ORT session run time: \(Float(endTime.uptimeNanoseconds - startTime.uptimeNanoseconds) / 1.0e6) ms")
```
YUNQIUGUO, line 76: Is the inference time info intended to be displayed? If so, we can use `Inference Time: X ms` to be consistent with the other sample applications.
(didn't see `inference time` is included in the screenshot though).

mobile/examples/speech_recognition/ios/readme.md
YUNQIUGUO: also development id for code sign requirement.

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    guard let output = outputs["output"] else {
+      throw SpeechRecognizerError.Error("Failed to get model output.")
+    }
+
+    let outputData = try output.tensorData() as Data
+    let result = outputData.withUnsafeBytes { (buffer: UnsafeRawBufferPointer) -> String in
```
YUNQIUGUO, line 83: do we need to use `guard` here?

mobile/examples/speech_recognition/ios/Podfile
```diff
@@ -0,0 +1,9 @@
+platform :ios, '11.0'
+
+target 'SpeechRecognition' do
+  use_frameworks!
+
+  #pod 'onnxruntime-mobile-objc', '1.8.1'
```
edgchen1, line 6: TODO use the right version

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
edgchen1: I don't think so, this is not an early return but a call with a closure that returns a result

mobile/examples/speech_recognition/ios/SpeechRecognition/AudioRecorder.swift
edgchen1: extra debugging info, will remove it

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
edgchen1: it's not intended to be displayed now, just a debug print

mobile/examples/speech_recognition/ios/readme.md
edgchen1: added some info to the readmes

skottmckay:
> 

Can we specify min versions for these?

---
Refers to: mobile/examples/basic_usage/model/requirements.txt:3 in 98ce006. [](commit_id = 98ce00655c0af459f583c8d180ed446eb3faf933, deletion_comment = False)

mobile/examples/speech_recognition/ios/readme.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### Install the Pod dependencies
+
+From this directory, run:
+
+```bash
+pod install
+```
+
```
skottmckay, line 38: Do we have a page anywhere that provides info on what you need to do if you need/want to switch from using the pre-built package to a custom build? 

I expect everyone new will use the pre-built package for simplicity, but assumably there will be some users who will want to go the custom build route. I guess we could always create a task and add that when someone asks.

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import Foundation
+
+private let kTokens = [
+  "<s>", "<pad>", "</s>", "<unk>", "|", "E", "T", "A", "O", "N", "I", "H", "S", "R", "D", "L", "U", "M", "W", "C", "F",
+  "G", "Y", "P", "B", "V", "K", "'", "X", "J", "Q", "Z",
+]
+
```
skottmckay, line 10: Comment explaining what these are would help. Is there existing documentation somewhere we could link to that explains the various values and how they're used?

mobile/examples/speech_recognition/ios/readme.md
edgchen1: not yet, as far as I know. I think we still need to figure out that process, whether it's building custom pods or using the custom build output more directly.

skottmckay (approved):
:shipit:

mobile/examples/speech_recognition/ios/SpeechRecognition/SpeechRecognizer.swift
edgchen1: added a comment but couldn't find a good documentation link. linked to the source example though.

PR #12, opened 2021-07-13
Add samples from the onnx runtime main repo

snnn:
These files are copied from https://github.com/microsoft/onnxruntime/tree/master/samples


iOS/ModelRunner/AppDelegate.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
edgchen1, line 1: FYI, we have mobile examples (including ios) under the mobile directory.
perhaps these examples should use the iOS packages too.

snnn: You mean, combine some of the [mobile/examples](https://github.com/microsoft/onnxruntime-inference-examples/tree/main/mobile/examples) and the iOS dir?

edgchen1: and perhaps adapt these examples to use the onnxruntime-mobile-c or onnxruntime-mobile-objc packages

snnn: I will create work items for them.

edgchen1:
will CI builds be added for all these samples?

snnn:
> will CI builds be added for all these samples?
> it would be good to verify that the samples are in working condition before adding them.

I added a Windows ci for the C/C++ code.  And thank you for the advice: I found we forgot to include some header files in our release package. So I commented out one of the build.  I will go back to change onnx runtime's packaging lines but it will take much time. So I would suggest we check in this change first and continue improving it.

edgchen1:
> > will CI builds be added for all these samples?
> > it would be good to verify that the samples are in working condition before adding them.
> 
> I added a Windows ci for the C/C++ code. And thank you for the advice: I found we forgot to include some header files in our release package. So I commented out one of the build. I will go back to change onnx runtime's packaging lines but it will take much time. So I would suggest we check in this change first and continue improving it.

sounds good. could we track which samples still need to be verified with TODO comments or somewhere else?

.github/workflows/win-ci.yml
```diff
@@ -0,0 +1,28 @@
+name: GitHub Actions Demo
```
edgchen1, line 1: update name

snnn:
model-explorer. Our packages miss experimental_onnxruntime_cxx_api.h and the example needs it. The author of the example authored the header file, but he  knew little about our packaging. I put a notes in c_cxx\CMakeLists.txt

iOS/README.md
```diff
@@ -0,0 +1,10 @@
+## ONNXRuntime iOS Sample
```
edgchen1, line 1:
```suggestion
# **WARNING: THIS EXAMPLE HAS NOT YET BEEN VERIFIED TO WORK WITH THE LATEST VERSION OF ONNX RUNTIME.**

## ONNXRuntime iOS Sample 
```

consider adding an obvious warning like the one above for examples which have yet to be verified and covered by CI tests.
perhaps also here:
swift/readme.md
python/readme.md (doesn't exist yet)
c_sharp/readme.md (doesn't exist yet)

edgchen1:
is this file used?
c_cxx/vs.png

iOS/README.md
gwang-msft: Maybe we can move all these API examples under a directory, say API_examples

snnn: @gwang-msft , you mean all the samples I'm copying in?  I can do it. @edgchen1, what do you think?

edgchen1: I don't have a strong opinion on whether it's in a subdirectory or not. Regarding the name "API_examples", the other examples show example API usage too. Is there a more specific name?

snnn: So, what are API examples and what are not? Sorry, what are the existing samples for in this repo?

gwang-msft: I don't have a strong opinion either, other examples here usually have a specific target platform, such as web or mobile, the ones added in the PR are more like examples on how to use a certain API such C++, C#

swift/ReadMe.md
```diff
@@ -0,0 +1,54 @@
+# MNIST Sample - Number recognition
+
+This sample uses the MNIST model from the Model Zoo: https://github.com/onnx/models/tree/master/vision/classification/mnist
+
+![Screenshot](Screenshot.png)
```
gwang-msft, line 5: I guess this screenshot was never checked in, maybe remove it?

jywu-msft (approved):

PR #20, opened 2021-08-04
Add a pipeline for compliance

snnn:
Add a pipeline for compliance

### DIFF
--- a/ci_build/azure_pipelines/compliance.yml
+++ b/ci_build/azure_pipelines/compliance.yml
@@ -8,7 +8,11 @@ jobs:
     displayName: 'Use Python 3.8.x'
 
   - powershell: |
-     pip --disable-pip-version-check install -r $(Build.SourcesDirectory)\mobile\examples\speech_recognition\model\requirements.txt
+     &C:\vcpkg\vcpkg.exe install protobuf:x64-windows-static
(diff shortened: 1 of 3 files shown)


ci_build/azure_pipelines/compliance.yml
```diff
@@ -0,0 +1,50 @@
...
+  pool:
+    vmImage: "windows-2019"
+  steps:
+  - task: UsePythonVersion@0
+    displayName: 'Use Python 3.8.x'
+    inputs:
+      versionSpec: 3.8.x
+
+  - powershell: |    
+     python -m pip --disable-pip-version-check install -r $(Build.SourcesDirectory)\mobile\examples\speech_recognition\model\requirements.txt
```
gwang-msft, line 13: Do we need to install pytorch? How about use `onnxruntime-inference-examples/mobile/examples/basic_usage/model/requirements.txt`?

gwang-msft (approved):

ci_build/azure_pipelines/compliance.yml
snnn: I don't know. Probably it doesn't matter much. We can revisit later. This pipeline is just for static code analysis.

PR #36, opened 2021-10-01
Add inference and accuracy evaluation for QDQ e2e example

chilo-ms:
Add inference and accuracy evaluation (exact match and f1 metrics) for QDQ BERT squad model.

### DIFF
--- a/quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
+++ b/quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
@@ -166,12 +166,13 @@ if __name__ == '__main__':
     print("QDQ model is saved to ", qdq_model_path)
 
     # QDQ model inference and get SQUAD prediction 
+    os.environ["ORT_TENSORRT_FP16_ENABLE"] = "1"  # Enable TRT FP16 precision
+    os.environ["ORT_TENSORRT_INT8_ENABLE"] = "1"  # Enable TRT INT8 precision
     session = onnxruntime.InferenceSession(qdq_model_path, providers=["TensorrtExecutionProvider", "CUDAExecutionProvider"])
 
     examples = dp.read_squad_json(squad_json)
     tokenizer = tokenization.BertTokenizer(vocab_file=vocab_file, do_lower_case=True)
     all_predictions = collections.OrderedDict()
-
     max_seq_length = sequence_lengths[-1] 
     doc_stride = 128
     max_query_length = 64
--- a/quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
+++ b/quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
@@ -11,10 +11,12 @@ import six
 import unicodedata
 import onnx
 import onnxruntime
+import sys
 import data_processing as dp
 import tokenization
 from pathlib import Path
 from onnxruntime.quantization import CalibrationDataReader, create_calibrator, write_calibration_table, QuantType, QuantizationMode, QLinearOpsRegistry, QDQQuantizer
+from squad.evaluate_v1_1 import evaluate 
 
 class BertDataReader(CalibrationDataReader):
     def __init__(self, model_path, squad_json, vocab_file, batch_size, max_seq_length, num_inputs):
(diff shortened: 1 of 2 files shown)


stevenlix:
There is a newer version evaluate-v2.0.py. Is it better to use it?

quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
```diff
@@ -63,11 +65,48 @@ def get_next(self):
                 input_ids = np.expand_dims(features[0].input_ids, axis=0)
                 input_mask = np.expand_dims(features[0].input_mask, axis=0)
                 segment_ids = np.expand_dims(features[0].segment_ids, axis=0)
-        data = [{"input_ids": input_ids, "input_mask": input_mask, "segment_ids":segment_ids}]
+        data = [{"input_ids": input_ids, "attention_mask": input_mask, "token_type_ids":segment_ids}]
         self.current_index += self.batch_size
         self.enum_data_dicts = iter(data)
         return next(self.enum_data_dicts, None)
 
+def inference(ort_session, features, example):
```
stevenlix, line 73: Can we make inference and evaluation more structural? for example, reuse BertDataReader to read data and define BertEvaluator for evaluation, just like what we did in other e2e examples.

chilo-ms: The reason I didn't use BertDataReader for inference and evaluation is that for some SQUAD examples, they might have more than one feature. Our current BertDataReader only considers one feature even though the example has multiple features. If we want to handle all the features for all the examples and at the same time run it in batch, we might need to modify the BertDataReader.

chilo-ms:
> There is a newer version evaluate-v2.0.py. Is it better to use it?

I can try to add evaluate-v2.0.py

quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
stevenlix: It's good to handle all the features. One thing is we need to make sure calibration are using same amount of features as inference/evaluation, otherwise calibration may not be able to fully represent inference data

chilo-ms: For evaluation, I think we must handle all the features of each example in order to achieve better accuracy. Therefore, current data reader should be modified for calibration and evaluation to handle all the features.

stevenlix: Yes. we should not use same data for both calibration and evaluation. As we did in other e2e model, we should split the data set, i.e. 80% vs 20%, for each. But the key is to use same pre-processing and data/feature handling for both.

chilo-ms:
> > There is a newer version evaluate-v2.0.py. Is it better to use it?
> 
> I can try to add evaluate-v2.0.py

evaluate-v2.0.py added as an option.

quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
```diff
@@ -125,4 +243,32 @@ def get_next(self):
...
+    sess_options.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_DISABLE_ALL
+    session = onnxruntime.InferenceSession(qdq_model_path, sess_options=sess_options, providers=["TensorrtExecutionProvider", "CUDAExecutionProvider"])
+    all_predictions = inference(data_reader, session) 
+
+    prediction_file = "./prediction.json"
+    with open(prediction_file, "w") as f:
+        f.write(json.dumps(all_predictions, indent=4))
+        print("\nOutput dump to {}".format(prediction_file))
+
+    # Evaluate QDQ model for SQUAD v1.1
```
stevenlix, line 263: why do we run evaluation in the separate script? Is it better to show evaluation result right after the inference is done since it's an e2e example?

chilo-ms: Yes, we can show result after the inference is done.
Will modify to execute these evalute-XXX.py directly inside our e2e example script.

drcege:
Hi, could you guys add the exact model ID or link to the used huggingface models? There are too many bert models to choose. I tried several but the output nodes did not match.

PR #42, opened 2021-10-12
Refine BERT data reader to avoid OOM when doing dataset pre-processing

chilo-ms:
We can use one data reader to do data pre-processing for the whole dataset, however,
some machines don't have sufficient memory to hold all dataset and all intermediate output,
especially using 'Entropy' or 'Percentile' calibrator which collects histogram for tensors.
So let multiple data readers to handle different stride of dataset to avoid OOM.

Also, made some changes to the variables in order to use the correct setting of quantization tool.


quantization/nlp/bert/trt/e2e_tensorrt_bert_example.py
```diff
@@ -207,19 +207,30 @@ def inference(data_reader, ort_session):
     vocab_file = "./squad/vocab.txt"
     augmented_model_path = "./augmented_model.onnx"
     qdq_model_path = "./qdq_model.onnx"
-    sequence_lengths = [384]
-    calib_num = 100
-    op_types_to_quantize = ['MatMul', 'Transpose', 'Add']
+    sequence_lengths = [128]
+    calib_num = 10
```
stevenlix, line 211: Is 10 enough to get good calibration results?

jywu-msft: 10 does seem too few.

chilo-ms: Yes, it's a typo. Made it back to 100.

stevenlix (approved):

PR #48, opened 2021-10-26
Update example model download URL to Azure storage static website.

edgchen1:
Moving files to a static website: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-static-website


shahicasper (commented):
h

PR #52, opened 2021-11-05
Add HuggingFace transformer dataloader as an option for e2e BERT QDQ example

chilo-ms:
We have implemented dataset processing, data loader and pre-post processing based on [Nvidia TensorRT demo BERT repo](https://github.com/NVIDIA/TensorRT/tree/master/demo/BERT) for PTQ QDQ. However, when doing evaluation on QDQ BERT large model, we still find F1 accuracy is slight lower (1 point) compared with using HuggingFace transformers data loader and pre-post processing. 
Therefore, we include HuggingFace transformers data loader and pre-post processing as an option for user to choose.


chilo-ms:
We don't include HF data loader until we know the difference between HF's and ours and can simplify it.

PR #82, opened 2022-02-23
Python API sample

natke:
Add Python API sample to support docs.

### DIFF
--- a/python/api/onnxruntime-python-api.py
+++ b/python/api/onnxruntime-python-api.py
@@ -4,7 +4,7 @@ import torch
 import onnxruntime
 
 MODEL_FILE = '.model.onnx'
-DEVICE_NAME = 'cpu' # Replace this with 'cuda' or equivalent for other devices
+DEVICE_NAME = 'cuda' if torch.cuda.is_available() else 'cpu'
 DEVICE_INDEX = 0     # Replace this with the index of the device you want to run on
 DEVICE=f'{DEVICE_NAME}:{DEVICE_INDEX}'
 
--- a/python/api/onnxruntime-python-api.py
+++ b/python/api/onnxruntime-python-api.py
@@ -118,9 +118,9 @@ def main():
     # tensor([0.7023, 1.3127, 1.7289, 0.3982, 0.8386])
 
     create_model(torch.int64)
-
+ 
     print(run_with_torch_tensors_on_device(torch.ones(5, dtype=torch.int64).to(DEVICE), torch.zeros(5, dtype=torch.int64).to(DEVICE), np_type=np.int64, torch_type=torch.int64))
-    # tensor([0.7023, 1.3127, 1.7289, 0.3982, 0.8386])
+    # tensor([1, 1, 1, 1, 1])
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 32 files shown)


python/api/onnxruntime-python-api.py
```diff
@@ -0,0 +1,117 @@
+# A set of code samples showing different usage of the ONNX Runtime Python API
```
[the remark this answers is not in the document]
natke: Do you have an example I can copy?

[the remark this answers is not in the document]
pranavsharma:
```
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License.
```

pranavsharma, line 1: Can also refer to C:\Users\prs\work_projects\onnxruntime\onnxruntime\test\python\onnxruntime_test_python_iobinding.py and 
C:\Users\prs\work_projects\onnxruntime\onnxruntime\test\python\onnxruntime_test_python.py

python/api/onnxruntime-python-api.py
```diff
@@ -0,0 +1,117 @@
...
+import numpy as np
+import torch
+import onnxruntime
+
+MODEL_FILE = '.model.onnx'
+DEVICE_NAME = 'cpu' # Replace this with 'cuda' or equivalent for other devices
```
pranavsharma, line 7: Using 'cuda' would be a better example since CPU to CPU copies done inside ```run_with_data_copied_to_device``` can confuse users to think they're required to do so.

natke: I guess I was hoping to demonstrate the functionality without users having to install CUDA. I can change it to CUDA though.

python/api/onnxruntime-python-api.py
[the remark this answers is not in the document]
natke: Added

natke: I'm hoping these can be self contained samples

python/api/onnxruntime-python-api.py
natke: Will set CUDA if it is available

python/api/onnxruntime-python-api.py
```diff
@@ -0,0 +1,127 @@
...
+# Run the model on CPU consuming and producing numpy arrays 
+def run(x: np.array, y: np.array) -> np.array:
+    session = create_session(MODEL_FILE)
+
+    z = session.run(["z"], {"x": x, "y": y})
+    
+    return z[0]   
+
+# Run the model on device consuming and producing ORTValues
+def run_with_data_copied_to_device(x: np.array, y: np.array) -> onnxruntime.OrtValue:
```
hariharans29, line 49: Nit: Can this be `run_with_iobinding` ? Or is the idea to be explicit about what is being done ? One tiny pedantic inaccuracy with this naming is that when this model runs on CPU (when CUDA is not available), then no "data copy to device" occurs, does it ?

hariharans29 (approved):

python/api/onnxruntime-python-api.py
natke: Would it be better named run_with_data_on_device()? (Then it is copied if GPU and already on device if CPU)

PR #90, opened 2022-03-12
Add another c++ simple example (Updated)

chilo-ms:
This PR is an update for the previous PR https://github.com/microsoft/onnxruntime-inference-examples/pull/87, due to some changes from Azure Pipelines, therefore this example folder needs to be updated as well.

### DIFF
--- a/c_cxx/squeezenet/CMakeLists.txt
+++ b/c_cxx/squeezenet/CMakeLists.txt
@@ -2,14 +2,14 @@ project(capi_test)
 
 set(CMAKE_BUILD_TYPE Debug)
 
-cmake_minimum_required (VERSION 2.6)
+cmake_minimum_required(VERSION 3.13)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include_directories( 
-    ${PROJECT_SOURCE_DIR}/../../include/onnxruntime/core/session/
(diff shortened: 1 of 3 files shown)


c_cxx/squeezenet/CMakeLists.txt
```diff
@@ -2,14 +2,14 @@ project(capi_test)
...
-cmake_minimum_required (VERSION 2.6)
+cmake_minimum_required(VERSION 3.13)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include_directories( 
-    ${PROJECT_SOURCE_DIR}/../../include/onnxruntime/core/session/
-    ${PROJECT_SOURCE_DIR}/../../include/onnxruntime/core/providers/tensorrt/
+    $ENV{ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/session/
```
snnn, line 11: It can be a cmake option instead of environment variable. For example, onnxruntime_CUDA_HOME is such a thing.

snnn: Indeed we already have such a thing: https://github.com/microsoft/onnxruntime-inference-examples/blob/main/c_cxx/CMakeLists.txt#L23

snnn: Usage: 
```
cmake .. -A x64 -T host=x64  -DONNXRUNTIME_ROOTDIR=c:\dev\ort_install
```

PR #91, opened 2022-03-15
Add squeezenet to CI build

snnn:
1. Add the squeezenet example to CI build
2. Validate all C/C++ examples with the latest onnxruntime nightly build. (Something is better than nothing)

### DIFF
--- a/c_cxx/squeezenet/main.cpp
+++ b/c_cxx/squeezenet/main.cpp
@@ -5,6 +5,8 @@
 #include <assert.h>
 #include <vector>
 #include <onnxruntime_cxx_api.h>
+
+#ifdef HAVE_TENSORRT_PROVIDER_FACTORY_H
 #include <tensorrt_provider_factory.h>
 #include <tensorrt_provider_options.h>
 
@@ -31,6 +33,7 @@ std::unique_ptr<OrtTensorRTProviderOptionsV2> get_default_trt_provider_options()
 
   return tensorrt_options;
 }
+#endif
 
 void run_ort_trt() {
   Ort::Env env(ORT_LOGGING_LEVEL_WARNING, "test");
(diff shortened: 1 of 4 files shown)


c_cxx/squeezenet/main.cpp
```diff
@@ -5,6 +5,8 @@
 #include <assert.h>
 #include <vector>
 #include <onnxruntime_cxx_api.h>
+
+#ifdef HAVE_TENSORRT_PROVIDER_FACTORY_H
 #include <tensorrt_provider_factory.h>
 #include <tensorrt_provider_options.h>
```
chilo-ms, line 11: Quick question here.
In line 41, the OrtTensorRTProviderOptionsV2 is defined in tensorrt_provider_options.h
It seems line 41 also should be in #ifdef HAVE_TENSORRT_PROVIDER_FACTORY_H ?

snnn: Interesting. Why it could work now?

snnn: SessionOptionsAppendExecutionProvider_TensorRT_V2 is declared in onnxruntime_c_api.h, it works similar. onnxruntime_c_api.h doesn't include tensorrt_provider_options.h.  So I think it is ok.

snnn: Oh, in onnxruntime_c_api.h we have "ORT_RUNTIME_CLASS(TensorRTProviderOptionsV2);" . So it is enough.

snnn:
@chilo-ms , there are a few warnings, I hope you can help fix: https://github.com/microsoft/onnxruntime-inference-examples/runs/5559177059?check_suite_focus=true

chilo-ms:
> @chilo-ms , there are a few warnings, I hope you can help fix: https://github.com/microsoft/onnxruntime-inference-examples/runs/5559177059?check_suite_focus=true

Okay, I will fix the warnings.

chilo-ms (approved):

PR #92, opened 2022-03-17
Update data_reader.py

mkmsftgit:
Class methods that access class variables should have "self." prefix.


snnn:
This PR has conflicts that must be resolved

PR #95, opened 2022-03-29
Support using full ORT iOS package in mobile examples

edgchen1:
Add support for using full ORT iOS package in mobile examples and update CI builds.


snnn (approved):

mobile/examples/object_detection/ios/ORTObjectDetection/prepare_model.requirements.txt
```diff
@@ -1,3 +1,4 @@
-onnxruntime
+onnx==1.10.0
+onnxruntime==1.10.0
 tensorflow==2.6.3
```
skottmckay, line 3: If I checkout ORT v1.10 is has ONNX v1.11 in cmake/external/onnx

PR #103, opened 2022-04-19
AzureML sample for questioning answering with ONNX Runtime

natke:
Add AzureML sample for deploying an ONNX model with ONNX Runtime, using latest AzureML API.

Direct link to the notebook: https://github.com/natke/onnxruntime-inference-examples/blob/azureml-sample/python/azureml/Inference_Bert_with_OnnxRuntime_on_AzureML.ipynb

(Based on this notebook, which will be deprecated after this sample is merged: https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/python/tools/transformers/notebooks/Inference_Bert_with_OnnxRuntime_on_AzureML.ipynb)


bstollnitz:
This is a great demo, Natalie! The Azure ML part seems solid to me.
Just a tiny nitpick: the comment in the "init()" function spells "intialization" instead of "initialization".

python/azureml/score.py
```diff
@@ -0,0 +1,113 @@
...
+            if tokens[i][0:2] == "##":
+                answer += tokens[i][2:]
+            else:
+                answer += " " + tokens[i]
+        results['answer'] = answer.capitalize()
+    else:
+        results['error'] = "I am unable to find the answer to this question. Can you please ask another question?"
+    return results
+
+# Perform the one-off intialization for the prediction. The init code is run once when the endpoint is setup.
```
bstollnitz, line 38: Tiny nitpick: "initialization" is spelled incorrectly.

bstollnitz (approved):

python/azureml/yml/env.yml
```diff
@@ -0,0 +1,23 @@
...
+# be automatically provisioned for runs with userManagedDependencies=False.
+
+# Details about the Conda environment file format:
+# https://conda.io/docs/user-guide/tasks/manage-environments.html#create-env-file-manually
+
+name: project_environment
+dependencies:
+  # The python interpreter version.
+  # Currently Azure ML only supports 3.5.2 and later.
+- python=3.6.2
```
tianleiwu, line 11: latest onnxruntime package does not support python 3.6. I recommend python 3.7 or 3.8.
But it is optional since older version of onnxruntime already has optimizations for BERT model.

python/azureml/yml/env.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- pip:
+  - numpy
+  - onnxruntime
```
tianleiwu, line 15: This environment seems for CPU only. For GPU environment, need onnxruntime-gpu package.

tianleiwu (approved):

python/azureml/score.py
natke: Fixed

PR #107, opened 2022-05-12
Add C# sample running an augmented model

natke:
Sample for generating an augmented model and running it in C#


wenbingl:
Nice pictures, is there any copyright issue?

wenbingl (approved):

natke:
> Nice pictures, is there any copyright issue?

I think it should be fine. The original model is in the zoo, and the augmented one is generated by the sample.

c_sharp/image_classification/model/augment_model.py
```diff
@@ -0,0 +1,20 @@
...
+
+mobilenet_model = onnx.load_model('mobilenetv2-12.onnx')
+augmented_model = pnp.SequentialProcessingModule(
+    pnp.PreMobileNet(224),
+    mobilenet_model,
+    pnp.PostMobileNet())
+
+# The image size is dynamic, the 400x500 here is to get a fake input to enable export
+fake_image_input = torch.ones(500, 400, 3).to(torch.uint8)
+augmented_model.forward(fake_image_input)
```
skottmckay, line 13: This is slightly misleading as it's not channels last, Something like 3, 500, 400 would be clearer.

c_sharp/image_classification/model/augment_model.py
```diff
@@ -0,0 +1,20 @@
+import onnx
+import torch
+from onnxruntime_extensions import pnp
+
+mobilenet_model = onnx.load_model('mobilenetv2-12.onnx')
+augmented_model = pnp.SequentialProcessingModule(
```
skottmckay, line 6: Where does this model come from? I wasn't part of the PR (binary files need to be manually added).

c_sharp/image_classification/README.md
```diff
@@ -0,0 +1,51 @@
+# Image classification using an augmented ONNX model
+
+This samples uses the utilities in [ONNX Runtime extensions](https://github.com/microsoft/onnxruntime-extensions) to augment the MobileNetV2 model to include the data pre and post processing operations required by the model.
```
skottmckay, line 3: >This samples uses the utilities in [ONNX Runtime extensions](https://github.com/microsoft/onnxruntime-extensions) to augment the MobileNetV2 model to include the data  [](http://example.com/codeflow?start=0&length=167)

nit: This sample...

I think this should also call out it's the PyTorch mobilenet (maybe with a link to it) as the pre-processing is specific to how that model was trained. 

TF does not use per-channel normalization so the pre-processing would be incorrect if it was TF mobilenet.

c_sharp/image_classification/README.md
```diff
@@ -0,0 +1,51 @@
...
+A simple C# console application to make predictions with the augmented model is contained in this directory.
+
+## Pre-requisites
+
+* Install cmake
+* Install gcc for your platform
+* Install conda
+* Create a conda environment
+* Install onnxruntime-extension package from source: python -m pip install git+https://github.com/microsoft/onnxruntime-extensions.git
+
```
skottmckay, line 16: Should this list all the other pip modules required, like 'torch'?

c_sharp/image_classification/model/augment_model.py
```diff
@@ -0,0 +1,20 @@
...
+    pnp.PreMobileNet(224),
+    mobilenet_model,
+    pnp.PostMobileNet())
+
```
skottmckay, line 10: Some comments in the code explaining what each line is doing, or link to doco on the ort-extensions repo, would help a user understand what is going on.

PR #108, opened 2022-05-16
Fix link to ONNX to ORT format model conversion info.

edgchen1:
Fix link to ONNX to ORT format model conversion info.

See https://github.com/microsoft/onnxruntime/issues/11519


YUNQIUGUO (approved):
thanks for updating the link!

PR #110, opened 2022-05-18
[docs] add link to doc explaining how to replace use of the pre-built package in samples with a custom build

YUNQIUGUO:
As title.

This PR just adds link to direct to Custom Build docs for convenience if user directly goes to the example repo.

Main updates are done in this pr [[docs] Add documentation about Android custom build](https://github.com/microsoft/onnxruntime/pull/11207)


edgchen1:
the samples aren't guaranteed to work with arbitrary custom builds. if we want to add custom build instructions to the existing samples (or perhaps a separate sample just for demonstrating how to use a custom build), perhaps we should also include op config and build settings files that would work for the sample.

natke:
What's the status of this PR? Can we merge it, or does it need more work?? @YUNQIUGUO

YUNQIUGUO:
> What's the status of this PR? Can we merge it, or does it need more work? @YUNQIUGUO

I guess we can close it for now. As if we want to demonstrate the usage of a custom-prebuilt package instead of a pre-built one, we might need to create another sample for that.

PR #128, opened 2022-07-26
Add MAUI example for mobile targets

skottmckay:
Largely replicate the Xamarin example with Windows, iOS and Android targets using MAUI.

Note: Currently requires Visual Studio 2022 Preview for .net 6 support.


mobile/examples/Maui/MauiVisionSample/MauiVisionSample/Models/Mobilenet/MobilenetSample.cs
```diff
@@ -0,0 +1,73 @@
...
+            // Run inference
+            using IDisposableReadOnlyCollection<DisposableNamedOnnxValue> results = Session.Run(inputs);
+
+            // Postprocess to get softmax vector
+            IEnumerable<float> output = results.First().AsEnumerable<float>();
+            float sum = output.Sum(x => (float)Math.Exp(x));
+            IEnumerable<float> softmax = output.Select(x => (float)Math.Exp(x) / sum);
+
+            // Extract top 3 predicted classes
+            IEnumerable<MobilenetPrediction> top10 = softmax
```
edgchen1, line 61: nit: the name `top10` doesn't match the top 3 values it contains

mobile/examples/Maui/MauiVisionSample/MauiVisionSample/VisionSampleBase.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            _prevAsyncTask = Initialize();
+            return _prevAsyncTask;
+        }
+
+        public async Task<ImageProcessingResult> ProcessImageAsync(byte[] image)
+        {
+            await AwaitLastTaskAsync().ConfigureAwait(false);
+
+            return await OnProcessImageAsync(image);
```
edgchen1, line 77: why is _prevAsyncTask not set here?

edgchen1:
are there instructions for how to build it? also, maybe not in this PR, but it would be good to have a CI build that builds this sample.

mobile/examples/Maui/MauiVisionSample/MauiVisionSample/VisionSampleBase.cs
skottmckay: We use the _prevAsyncTask when we're re/creating the InferenceSession. As that can be slow we don't actually `await` that operation here (we just save the Task for it in _prevAsyncTask and return) so it can continue in the background. 

We do `await` the OnProcessImageAsync here so I believe it's treated differently because of that.

skottmckay:
> are there instructions for how to build it? also, maybe not in this PR, but it would be good to have a CI build that builds this sample.

Not sure how feasible that would be. Separate repo to our usual CIs, would probably need AppCenter to be able to test, to build the iOS app you need a mac device, and also requires a pre-release VS which is not available in the standard CI images.

mobile/examples/Maui/MauiVisionSample/readme.md
```diff
@@ -0,0 +1,105 @@
...
+As execution providers must be selected prior to the creation of the `InferenceSession` this necessitates an expensive re-creation of the session in the sample app. Typically you would pre-determine which execution providers you wanted enabled and create the session once with that information.
+
+The CPU Execution Provider will be able to run all models.
+
+In some scenarios it may be beneficial to use the NNAPI Execution Provider on Android, or the CoreML Execution Provider on iOS. 
+This is highly dependent on your model, and also the device (particularly for NNAPI), so it is necessary for you to performance test.
+
+## Image acquisition
+
+There are 3 potential ways to acquire the image to process in the sample app.
```
YUNQIUGUO, line 71: > There are 3 potential ways to acquire the image

nit: from info below, seems like there are 2 ways to acquire the image? ( 3 stages for image processing

skottmckay: The 3 I'm referring to are using the included sample image (GetSampleImageAsync), picking an existing image from the device (PickPhotoAsync), or taking a photo (TakePhotoAsync).

mobile/examples/Maui/MauiVisionSample/readme.md
```diff
@@ -0,0 +1,105 @@
...
+Classifies the major object in the image into 1,000 pre-defined classes.
+
+### [Ultraface](https://github.com/onnx/models/tree/f064171f7dd8e962a8a5b34eac8e1bcf83cebbde/vision/body_analysis/ultraface#ultra-lightweight-face-detection-model)
+
+Lightweight face detection model designed for edge computing devices providing detection boxes and scores for a given image.
+
+This model is included in the repository, but has been updated using the onnxruntime python package tools to:
+- remove unused initializers,
+  - `python -m onnxruntime.tools.optimize_onnx_model --opt_level basic <model>.onnx <updated_model>.onnx`
+- make the initializers constant
```
edgchen1, line 19: not obvious to me how the initializers are made constant with the command, I thought it updated the ONNX opset to 14?

skottmckay: Sorry - I included the wrong command. It was a previous opset update that causes the issue. Updated with correct info to fix the issue.

mobile/examples/Maui/MauiVisionSample/readme.md
```diff
@@ -0,0 +1,106 @@
+# ONNX Runtime MAUI Vision Sample
+
+The [MAUI Vision Sample](MauiVisionSample.sln) demonstrates the use of two different vision models from the [ONNX Model Zoo collection](https://github.com/onnx/models/tree/main), by a [MAUI](https://docs.microsoft.com/en-us/dotnet/maui/what-is-maui) app.
+
+## Overview
+The app enables you to take/pick a photo on the device or use a sample image to explore the following models.
+
+### [Mobilenet](https://github.com/onnx/models/tree/main/vision/classification/mobilenet)
```
edgchen1, line 8: nit: should we use a permalink here too, like the Ultraface link?

edgchen1 (approved):

edgchen1:
> > are there instructions for how to build it? also, maybe not in this PR, but it would be good to have a CI build that builds this sample.
> 
> Not sure how feasible that would be. Separate repo to our usual CIs, would probably need AppCenter to be able to test, to build the iOS app you need a mac device, and also requires a pre-release VS which is not available in the standard CI images.

maybe we can consider it when it doesn't require a pre-release VS. we do have some existing CI builds in this repo.

mobile/examples/Maui/MauiVisionSample/readme.md
skottmckay: I went the other way and changed the link for ultraface so that if new versions of the model are added (e.g. one with the fixes we are manually making) they will show up. 

Cost is that if the page moves completely the link will break, but that seems less likely. e.g. for mobilenet new versions were added relatively recently.

shahicasper: j

shahicasper (commented):
h

PR #134, opened 2022-08-19
add qdq debugging example

chenfucn:
Adding example run_qdq_debug.py

### DIFF
--- a/quantization/image_classification/cpu/run.py
+++ b/quantization/image_classification/cpu/run.py
@@ -52,6 +52,9 @@ def main():
     dr = resnet50_data_reader.ResNet50DataReader(
         calibration_dataset_path, input_model_path
     )
+
+    # Calibrate and quantize model
+    # Turn off model optimization during quantization
     quantize_static(
         input_model_path,
         output_model_path,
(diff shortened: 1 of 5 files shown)


quantization/image_classification/cpu/run.py
```diff
@@ -52,13 +52,17 @@ def main():
     dr = resnet50_data_reader.ResNet50DataReader(
         calibration_dataset_path, input_model_path
     )
+
```
yufenglee, line 55: call the pre-precess?

chenfucn: It would break the flow described in Readme.md if we call pre-processing here.

chenfucn: I think it might be better to make the pre-processing front and center. In the example, highlighting it to be a separated step. If I call it in her, it is kinda hidden. The user must dig into python code to locate it. In current layout, this step is on the command line level instead of python code level. This might be more obvious to the user. What do you think?

chenfucn:
rebased on the main

yufenglee (approved):

PR #135, opened 2022-08-22
Update MauiVisionSample SkiaSharp dependency version to 2.88.1.

edgchen1:


skottmckay (approved):
Would be good to mention in the checkin comment that this addresses the SkiaSharp issue.

YUNQIUGUO (approved):

PR #144, opened 2022-09-22
Remove deprecated API usage

yuslepukhin:
`GetInputName/GetOutputName` has been deprected last release and now removed.

### DIFF
--- a/c_cxx/MNIST/MNIST.cpp
+++ b/c_cxx/MNIST/MNIST.cpp
@@ -40,7 +40,8 @@ struct MNIST {
     const char* input_names[] = {"Input3"};
     const char* output_names[] = {"Plus214_Output_0"};
 
-    session_.Run(Ort::RunOptions{nullptr}, input_names, &input_tensor_, 1, output_names, &output_tensor_, 1);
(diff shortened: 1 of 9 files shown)


snnn (approved):

snnn:
I had a discussion with @yuslepukhin offline, we decided to merge this PR first despite it will cause pipeline failures. Because we want to unblock ORT release first. Otherwise we will stick in a deadlock.

jywu-msft:
when will we fix the pipeline failures? 
it caused me some confusion because I didn't know why it was failing at first. 
@hectorsvc FYI

yuslepukhin:
> when will we fix the pipeline failures? it caused me some confusion because I didn't know why it was failing at first. @HectorSVC FYI

I am open to suggestions. Apparently, with the current setup it will fix itself with the new release. From my conversation with with @snnn the samples are based on the previous release and that needs to be fixed.

snnn:
Ideally we should make our code be compatible with the latest ORT release and the latest ORT code in the main branch.  If we can't make it happen, I will modify the pipeline so that it would use a fixed version ORT nightly package instead.  Does it sound good?

PR #145, opened 2022-09-23
Fix build errors

pranavsharma:
- Fix include path for providers
- std::optional requires C++17
- std::unary_function is removed from C++17
Fixes https://github.com/microsoft/onnxruntime/issues/13058


c_cxx/include/providers.h
```diff
@@ -2,23 +2,23 @@
 // Licensed under the MIT License.
 
 #pragma once
-#include "cpu_provider_factory.h"
+#include "cpu/cpu_provider_factory.h"
```
snnn, line 5: In our nuget package and zip/tarball, all these headers are in the same folder. So "cpu/cpu_provider_factory.h" doesn't exist if the build directly fetch bits from our release page.

pranavsharma: Good point. However, it's our nuget pkg generation script that is copying the headers to a single include folder. At this point (last week of release) I'm hesitant to touch the core cmake files to install them elsewhere. The fastest option is to add the individual provider folders in the include path. At least customers won't have a bad experience running the samples. Would that work?

snnn:
The build failures are expected. I will refactor the pipelines to let them fetch bits from our ORT-Nightly feed to fix the issues.

c_cxx/CMakeLists.txt
```diff
@@ -31,7 +31,16 @@ if(NOT ONNXRUNTIME_ROOTDIR)
 endif()
 
 #TODO: we should only need one of them.
-include_directories("${ONNXRUNTIME_ROOTDIR}/include" "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/session")
+include_directories(
+  "${ONNXRUNTIME_ROOTDIR}/include/"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/cpu"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/cuda"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/dnnl"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/nuphar"
```
snnn, line 39: nuphar is deleted from the latest ORT.

c_cxx/CMakeLists.txt
```diff
@@ -31,7 +31,16 @@ if(NOT ONNXRUNTIME_ROOTDIR)
...
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/tensorrt"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/dml"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/providers/migraphx"
+  "${ONNXRUNTIME_ROOTDIR}/include/onnxruntime/core/session")
```
snnn, line 43: So, it means we need to list all EPs here? Why?

If I  understood correctly, ONNXRUNTIME_ROOTDIR is where ORT is installed, not the source folder. At the installation step, should we already put all the EP specific header files into one place?

PR #153, opened 2022-10-12
Update mobile samples to work with ORT 1.13

edgchen1:
The main change is that the ORT format was updated in a non-backwards-compatible way, so we need to use the current ORT version (1.13) to convert ONNX models to ORT format.


mobile/examples/object_detection/ios/ORTObjectDetection/prepare_model.requirements.txt
```diff
@@ -1,5 +1,8 @@
 onnx==1.10.0
-onnxruntime==1.10.0
+# TODO when it's released, use onnxruntime 1.13.0 instead of ort-nightly
```
edgchen1, line 2: use ORT 1.13 released package when available

snnn (approved):

shahicasper (commented):
f

PR #155, opened 2022-10-19
Update Android image classification example to use ONNX model zoo

edgchen1:
Motivation: Move away from downloading pre-converted ORT format models from Azure storage container.


mobile/examples/image_classification/android/prepare_models.requirements.txt
```diff
@@ -0,0 +1,2 @@
+onnx==1.12.0
+onnxruntime==1.12.1
```
edgchen1, line 2: if this PR is not merged before the ORT release, update to 1.13.

mobile/examples/image_classification/android/prepare_models.py
```diff
@@ -0,0 +1,102 @@
...
+        "https://raw.githubusercontent.com/pytorch/hub/master/imagenet_classes.txt",
+        args.output_dir / "imagenet_classes.txt",
+    )
+
+    # prepare models in temporary directory, then copy them to output directory
+    with tempfile.TemporaryDirectory(dir=args.output_dir) as temp_dir:
+        temp_dir = pathlib.Path(temp_dir)
+
+        for model_name, model_path_stem in [
+            ("MobileNet v2-1.0-int8", "mobilenetv2_int8"),
```
YUNQIUGUO, line 59: nit: Is it okay to use int8 mobilenet model? just noticed previously the model was an uint8 one.

edgchen1: it seems to work. I'm not sure how much of a difference it makes. @yufenglee any idea?
the ONNX model zoo only has an int8 quantized version.

yufenglee: I would recommend int8 model for mobile. Int8 will be faster than uint8 on arm64 without dot product.

mobile/examples/image_classification/android/prepare_models.requirements.txt
edgchen1: will do this in another PR

PR #163, opened 2022-10-25
Rewrite the fns_candy_style_transfer example so that it doesn't depends on libpng on Windows

snnn:
1. Rewrite the fns_candy_style_transfer example so that it doesn't depends on libpng on Windows
2. Fix a problem in pipeline's yaml file: {{parameters.ortversion}} was not evaluated because I forgot to put a `$` sign in front of it.


c_cxx/fns_candy_style_transfer/fns_candy_style_transfer.c
```diff
@@ -10,7 +9,10 @@
 #include "providers.h"
 #endif
 #include <objbase.h>
+#else
+#include <png.h>
```
edgchen1, line 13: do we need to include png.h here or is it enough to do so from image_file_libpng.c?

c_cxx/imagenet/runnable_task.h
```diff
@@ -3,9 +3,10 @@
 
 #pragma once
 #include <functional>
+
 #include "sync_api.h"
 
-class RunnableTask : public std::unary_function<void, void> {
+class RunnableTask {
```
edgchen1, line 9: after removing std::unary_function do we still need `#include <functional>`?

c_cxx/fns_candy_style_transfer/CMakeLists.txt
```diff
@@ -1,9 +1,14 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-add_executable(fns_candy_style_transfer "fns_candy_style_transfer.c")
+add_executable(fns_candy_style_transfer "fns_candy_style_transfer.c" "image_file.h")
+if(WIN32)
+  target_sources(fns_candy_style_transfer PRIVATE image_file_wic.cc)
+else()
+  target_sources(fns_candy_style_transfer PRIVATE image_file_libpng.c)
```
edgchen1, line 8: my understanding is that libpng things (PNG_LIBRARIES, PNG_INCLUDE_DIRS, PNG_LIBDIR) aren't used on windows. perhaps it would be clearer to put them in this block?

ci_build/download_ort_release.sh
```diff
@@ -0,0 +1,17 @@
...
+
+while getopts i: parameter_Option
+do case "${parameter_Option}"
+in
+i) ORT_VER=${OPTARG};;
+esac
+done
+
+apt-get update
+apt-get install -y cmake gcc g++ libpng-dev libjpeg-turbo8-dev curl
```
edgchen1, line 14: nit: this is not really part of "downloading ORT", maybe it could go in a separate script or this one could be renamed

c_cxx/fns_candy_style_transfer/image_file_wic.cc
```diff
@@ -0,0 +1,105 @@
...
+  wil::com_ptr_failfast<IWICBitmapDecoder> decoder;
+  FAIL_FAST_IF_FAILED(
+      piFactory->CreateDecoderFromFilename(input_file, NULL, GENERIC_READ,
+                                           WICDecodeMetadataCacheOnDemand,  // defer parsing non-critical metadata
+                                           &decoder));
+  UINT count = 0;
+  FAIL_FAST_IF_FAILED(decoder->GetFrameCount(&count));
+  if (count != 1) {
+    printf("The image has multiple frames, I don't know which to choose.\n");
+    abort();
```
edgchen1, line 30: why not `return -1`?

ci_build/download_ort_release.sh
snnn: Updated.

c_cxx/fns_candy_style_transfer/image_file_wic.cc
snnn: Thanks! I updated the code.

c_cxx/fns_candy_style_transfer/CMakeLists.txt
snnn: Yes, You are right.

edgchen1 (approved):

PR #172, opened 2022-11-29
Windows io buffer sample

sfatimar:
This change is to ensure that samples are valid for both dGPU and iGPU in a multi gpu system


sfatimar:
@jywu-msft Please review

sfatimar:
> I think it would be nice to provide some more details in the README to explain the differences between the two squeezenet samples. The user may not be familiar with "iobuffer optimization" and what it means. Would it be better to say one example is for cpu and the other one is for running on gpu's (and the optimization is there for memory copy avoidance?)

**We have made the change**

jywu-msft (approved):

PR #184, opened 2023-01-12
Update torch version in sample requirements.txt

YUNQIUGUO:
Update to a safer and newer torch version in sample requirements to solve the CG alert:

https://aiinfra.visualstudio.com/Lotus/_componentGovernance/97926/alert/8088465?typeId=13210915&pipelinesTrackingFilter=0

### DIFF
--- a/mobile/examples/speech_recognition/model/requirements.txt
+++ b/mobile/examples/speech_recognition/model/requirements.txt
@@ -1,6 +1,6 @@
 onnx==1.10.0
 onnxruntime==1.13.1
 protobuf==3.20.2
-torch==1.9.0
+torch==1.13.1
 torchaudio==0.9.0
 transformers==4.6.1
--- a/mobile/examples/speech_recognition/model/requirements.txt
+++ b/mobile/examples/speech_recognition/model/requirements.txt
@@ -2,5 +2,5 @@ onnx==1.10.0
 onnxruntime==1.13.1
 protobuf==3.20.2
 torch==1.13.1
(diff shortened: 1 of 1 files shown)


edgchen1:
torchaudio has an incompatible version

```
ERROR: Cannot install -r ../model/requirements.txt (line 5) and torch==1.13.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested torch==1.13.1
    torchaudio 0.9.0 depends on torch==1.9.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

YUNQIUGUO:
> torchaudio has an incompatible version
> 
> ```
> ERROR: Cannot install -r ../model/requirements.txt (line 5) and torch==1.13.1 because these package versions have conflicting dependencies.
> 
> The conflict is caused by:
>     The user requested torch==1.13.1
>     torchaudio 0.9.0 depends on torch==1.9.0
> 
> To fix this you could try to:
> 1. loosen the range of package versions you've specified
> 2. remove package versions to allow pip attempt to solve the dependency conflict
> ```

I see. um is it okay to update the torchaudio to the latest version 0.13.1 to solve the issue?

edgchen1:
> I see. um is it okay to update the torchaudio to the latest version 0.13.1 to solve the issue?

I don't see any issue with that - can try and see if it works.

edgchen1 (approved):

PR #185, opened 2023-01-12
[Mobile] Add super resolution ios sample app with ort-extensions support for ppp

YUNQIUGUO:
As title.

Used the pre-release version ort-extensions-c pod in the sample
TODO:  
- can update when available:
              1.  the path to register custom ops call for extensions later
              2. release version pod
              3. opset18 version super res model with better resolution results.
 - refine UI to add more options including integrate functions as camera capture input picture/pictures from photo gallery, etc. currently the sample app performs super res on displayed selected image from assets.


mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolution/ContentView.swift
```diff
@@ -0,0 +1,58 @@
...
+                    Image("cat_224x224").frame(width: 250, height: 250)
+                    
+                    Button("Perform Super Resolution") {
+                        performSuperRes.toggle()
+                    }
+                    
+                    if performSuperRes {
+                        Text("Output high resolution image: ").frame(width: 350, height: 40, alignment:.leading)
+                        
+                        if (runOrtSuperResolution() != nil) {
```
edgchen1, line 40: can we call runOrtSuperResolution() once and save the return value?

mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolution/ContentView.swift
```diff
@@ -0,0 +1,58 @@
...
+
+struct ContentView: View {
+    @State private var performSuperRes = false
+    
+    func runOrtSuperResolution() -> UIImage? {
+        do {
+            let outputImage = try ORTSuperResolutionPerformer.performSuperResolution()
+            return outputImage
+        } catch let error as NSError {
+            print("Error: \(error.localizedDescription)")
```
edgchen1, line 15: is it easy to see the error message or should we show it in the UI?

mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolution/ContentView.swift
```diff
@@ -0,0 +1,59 @@
...
+                    
+                    Button("Perform Super Resolution") {
+                        performSuperRes.toggle()
+                    }
+                    
+                    if performSuperRes {
+                        Text("Output high resolution image: ").frame(width: 350, height: 40, alignment:.leading)
+                        
+                        let outputImage = runOrtSuperResolution()
+                        if (outputImage != nil) {
```
edgchen1, line 41: can use swift optional binding:
```suggestion
                        if let outputImage = runOrtSuperResolution() {
```

mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolution/ContentView.swift
YUNQIUGUO: I guess all the print message in swift app will shown in the console output area. looks like it's common to use try/catch print(error message) in swift error handling.

currently the UI will show `unable to perform super resolution,` the specific error message should be easy to see in the console output area. but we can also consider to add it to UI of course.

edgchen1 (approved):
looks good.
doesn't have to be in this PR, but we should also consider building and testing this in the CI pipeline.

PR #187, opened 2023-01-23
Fix option name in QDQQuantizer

stevenlix:
OpTypesToExcludeOutputQuantizatioin should be OpTypesToExcludeOutputQuantization


jywu-msft (approved):
we need some tests to exercise these examples.

PR #195, opened 2023-01-30
Add super resolution example to mobile examples ci pipeline

YUNQIUGUO:
- Add super res android example to mobile examples ci pipeline
- Add test configuration to the super res ios xcodeproj and added commands to run in ci pipeline.

Can update to Testing with Full/Mobile packages as the other samples in the future, start with testing with the existing pre-release version pods for ios and checked-in aar package for android.

### DIFF
--- a/mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolutionTests/ORTSuperResolutionTests.swift
+++ b/mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolutionTests/ORTSuperResolutionTests.swift
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+import XCTest
+
+@testable import ORTSuperResolution
+
+final class ORTSuperResolutionTests: XCTestCase {
+
+    func testPerformSuperResolution() throws {
+        // test that it doesn't throw
+        try ORTSuperResolutionPerformer.performSuperResolution()
+    }
+}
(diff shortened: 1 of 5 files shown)


mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolutionTests/ORTSuperResolutionTests.swift
```diff
@@ -0,0 +1,15 @@
...
+import XCTest
+
+@testable import ORTSuperResolution
+
+final class ORTSuperResolutionTests: XCTestCase {
+
+    func testPerformSuperResolution() throws {
+        // test that it doesn't throw
+        try ORTSuperResolutionPerformer.performSuperResolution()
+    }
```
skottmckay, line 14: Is there any other testing we could do or is that validation covered by unit tests and this is just a sanity check where if it runs without throwing that's all we need to know.

YUNQIUGUO: The method `performSuperResolution()` contains steps including loadModel, create inference session, loadImage, registercustomops, etc already. and it will throw runtime error if any of the calls failed. 

I guess that's also what we do for simple unit testing for other sample apps, (i.e. test load model, create session, etc.)

https://github.com/microsoft/onnxruntime-inference-examples/blob/main/mobile/examples/super_resolution/ios/ORTSuperResolution/ORTSuperResolution/ORTSuperResolutionPerformer.mm#L24

skottmckay: It returns a UIImage* so could we check that's not null and has the expected height/width? 

Also performSuperResolution has a big try/catch and sets `error` if there's an exception. Given that, isn't it unlikely to throw and we at least should be checking the return value and `error` is not null ?

YUNQIUGUO: okay sorry I see. 

will update the tests here, and maybe consider to add separate UItesting scheme for this.

skottmckay (approved):

PR #201, opened 2023-02-09
[Mobile SuperRes] Update ort session creation to app onCreate() stage

YUNQIUGUO:
- Changes to avoid session creation happening with every instance instantiation.

- Other minor changes

### DIFF
--- a/mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
+++ b/mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
@@ -19,6 +19,7 @@ import java.util.concurrent.Executors
 
 class MainActivity : AppCompatActivity() {
     private var ortEnv: OrtEnvironment = OrtEnvironment.getEnvironment()
+    private lateinit var ortSession: OrtSession
     private var outputImage: ImageView? = null
     private var superResolutionButton: Button? = null
 
@@ -30,11 +31,16 @@ class MainActivity : AppCompatActivity() {
         outputImage = findViewById(R.id.imageView2);
         superResolutionButton = findViewById(R.id.super_resolution_button)
 
+        // Initialize Ort Session and register custom ops and register the onnxruntime extensions
+        // package that contains the custom operators.
+        // Note: These are used to decode the input image into the format the original model requires,
+        // and to encode the model output into png format
         val sessionOptions: OrtSession.SessionOptions = OrtSession.SessionOptions()
         sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+        ortSession = ortEnv.createSession(readModel(), sessionOptions)
 
         superResolutionButton?.setOnClickListener {
-            performSuperResolution(ortEnv.createSession(readModel(), sessionOptions))
+            performSuperResolution(ortSession)
             Toast.makeText(baseContext, "Super resolution performed!", Toast.LENGTH_SHORT).show()
         }
     }
@@ -42,6 +48,7 @@ class MainActivity : AppCompatActivity() {
     override fun onDestroy() {
         super.onDestroy()
         ortEnv.close()
+        ortSession.close()
     }
 
     private fun updateUI(result: Result) {
--- a/mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
+++ b/mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
@@ -31,8 +31,7 @@ class MainActivity : AppCompatActivity() {
         outputImage = findViewById(R.id.imageView2);
         superResolutionButton = findViewById(R.id.super_resolution_button)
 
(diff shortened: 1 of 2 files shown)


mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
```diff
@@ -30,8 +30,11 @@ class MainActivity : AppCompatActivity() {
         outputImage = findViewById(R.id.imageView2);
         superResolutionButton = findViewById(R.id.super_resolution_button)
 
+        val sessionOptions: OrtSession.SessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+
         superResolutionButton?.setOnClickListener {
-            performSuperResolution()
+            performSuperResolution(ortEnv.createSession(readModel(), sessionOptions))
```
natke, line 37: Trying to understand the scoping here: so the createSession() is called when the setOnClickListener is declared?

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/SuperResPerformer.kt
```diff
@@ -27,36 +26,32 @@ internal class SuperResPerformer(
...
+            output.use {
+                val rawOutput = (output?.get(0)?.value) as ByteArray
+                val outputImageBitmap =
+                    byteArrayToBitmap(rawOutput)
+
+                // Step 5: set output result
+                result.outputBitmap = outputImageBitmap
             }
         }
+        ortSession.close()
```
natke, line 49: Wouldn't we want the session to stay open for subsequent calls?

natke (commented):
Do we also need to make these changes for the iOS app?

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/SuperResPerformer.kt
skottmckay: Yeah - the examples should make it clear you want to create the session once and keep it around as long as you want to do any inferencing. It's an extremely expensive operation, and OrtSession.run is designed to be called multiple times including concurrently.

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
```diff
@@ -30,8 +30,11 @@ class MainActivity : AppCompatActivity() {
         outputImage = findViewById(R.id.imageView2);
         superResolutionButton = findViewById(R.id.super_resolution_button)
 
+        val sessionOptions: OrtSession.SessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+
```
skottmckay, line 35: As this is an example we should have comments explaining why we're doing things as the reader is unlikely to be familiar with ORT and this is how we can teach them what to do.

e.g. 'Register the onnxruntime extensions package that contains the custom operators. These are used to decode the input image into the format the original model requires, and to encode the model output into png format'

Can we also use the new 'registerCustomOpsUsingFunction' function so the examples are consistent across platforms?

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/SuperResPerformer.kt
YUNQIUGUO: > Do we also need to make these changes for the iOS app?

For the ios app, it's probably fine.  The inference session is only being created once.

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
YUNQIUGUO: sorry - didn't mean to put the create session here. will move it out of this, keep it in OnCreate() along with the registerCustomop steps.

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
YUNQIUGUO: Make sense. thanks for the suggestion. will add comments and notes here.

> Can we also use the new 'registerCustomOpsUsingFunction' function so the examples are consistent across platforms?

Regarding this, yes we should definitely update to this method for registering the custom ops. 

haven't get it tested yet (for iOS as well.) will update.

natke (approved):

mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/SuperResPerformer.kt
YUNQIUGUO: moved session.close() to `OnDestory(). ` thanks for the catch.

PR #202, opened 2023-02-14
Update ONNX to 1.13 due to security warning.

skottmckay:
Update ONNX to 1.13 due to security warning.

> Versions of the package onnx before 1.13.0 are vulnerable to Directory Traversal as the external_data field of the tensor proto can have a path to the file which is outside the model current directory or user-provided directory, for example "../../../etc/passwd"

Also update transformers to newer version that supports python 3.10. The existing version triggers and attempt to build the tokenizers python package dependency from source which requires the Rust compiler. Model generation was successful with the new version.


shahicasper (commented):
```

PR #203, opened 2023-02-14
[Mobile super res Android] Minor refinement to super res android application

YUNQIUGUO:
1. avoid copying test image to drawable
2. display output in one screen
3. avoid app crash when input image is not supported by model


mobile/examples/super_resolution/android/app/src/main/java/ai/onnxruntime/example/superresolution/MainActivity.kt
```diff
@@ -60,7 +72,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun readInputImage(): InputStream {
-        return assets.open("test_superresolution.png")
+        return assets.open("cat_224x224.png")
```
natke, line 75: Actually, can we leave the name as test_superresolution.png. Then instructions to use a different image will be to copy into this filename

natke (approved):

PR #210, opened 2023-02-21
Update mobile object detection model preparation script to use ONNX 1.13.

edgchen1:


YUNQIUGUO (approved):
onnx and protobuf dependency are not required anymore?

edgchen1:
> onnx and protobuf dependency are not required anymore?

they still are (as dependencies of these ones), but they are not specified explicitly now.

edgchen1:
from [build logs](https://aiinfra.visualstudio.com/PublicPackages/_build/results?buildId=288337&view=logs&j=16fca9d0-6d19-5b64-f2c0-4a873fa5acd3&t=d8015253-8cdc-5738-2a2a-a54d876a7eca):
```
ERROR: Cannot install -r ./prepare_model.requirements.txt (line 1), -r ./prepare_model.requirements.txt (line 2) and -r ./prepare_model.requirements.txt (line 3) because these package versions have conflicting dependencies.

The conflict is caused by:
    onnx 1.13.1 depends on protobuf<4 and >=3.20.2
    onnxruntime 1.14.1 depends on protobuf
    tensorflow 2.11.0 depends on protobuf<3.20 and >=3.9.2
```

the next version of tensorflow should hopefully have a compatible protobuf dependency:
https://github.com/tensorflow/tensorflow/blob/e0059bba8cb4d6d3979458c0945c66b4cc1c9f69/tensorflow/tools/pip_package/setup.py#L100

since the point of this PR is to update onnx to 1.13 and we also depend on tensorflow for this example, let's wait for tensorflow 2.12 and try that.

PR #214, opened 2023-03-09
[Android Demo] question answering

wejoncy:
1. Add CI of QuestionAnswering Android Sample
2. Add Sample of Android QuestionAnswering 

Model is Mobilebert from huggingface, and we will convert it to onnx model, followed by a quantization operation, and add pre-post-processing at last.


wejoncy:
Hi, @YUNQIUGUO looks like the script `prepare_model.py` can be shared between Android and IOS.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private fun readModel(): ByteArray {
+        val modelID = R.raw.csarron_mobilebert_uncased_squad_v2_quant_with_pre_post_processing
+        return resources.openRawResource(modelID).readBytes()
+    }
+
+    private fun readQuestion(): CharSequence {
+        var user_text = QuestionText?.text
+        return if (user_text?.isEmpty() == true) QuestionText?.hint!! else user_text!!
```
YUNQIUGUO, line 92: we might can eliminate the usage of nullable object type here to eliminate the many usage of `!!` force operator for converting nullable to its non-nullable equivalent here.

mobile/examples/question_answering/android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<resources>
+    <string name="app_name">ORT QuestionAnswering</string>
+    <string name="perform_qa">GO</string>
```
YUNQIUGUO, line 3: maybe a [->] button would make more sense then [GO] for perform QA?

as a reference: TFLite Q&A sample looks like this: https://github.com/tensorflow/examples/tree/master/lite/examples/bert_qa/android#tensorflow-lite-bert-question--answer-demo-application

mobile/examples/question_answering/android/README.md
```diff
@@ -0,0 +1,62 @@
...
+### Step 1: Clone the ONNX runtime mobile examples source code
+
+Clone this repository to get the sample application. 
+
+`git@github.com:microsoft/onnxruntime-inference-examples.git`
+
+
+### Step 2: Prepare the model and required packages
+
+- The required aar package is under lib's folder `mobile\examples\question_answering\android\app\libs`.
```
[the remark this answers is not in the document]
wejoncy: I need to make sure we have embedded ort-ext by default.

mobile/examples/question_answering/android/app/src/main/res/raw/csarron_mobilebert_uncased_squad_v2_quant_with_pre_post_processing.onnx
```diff
@@ -0,0 +1 @@
+model is generated by 'python ./prepare_models.py'
```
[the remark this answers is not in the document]
wejoncy: It's just a placeholder and can be removed if necessary.  Given that it does not account for too much space I would rather keep it here for file structure.

mobile/examples/question_answering/android/README.md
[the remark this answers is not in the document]
wejoncy: Yeah. We don't need it anymore.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
wejoncy: Not quite follow,  do you mean  "var user_text = QuestionText?.text!!"?

mobile/examples/question_answering/android/app/src/main/res/values/strings.xml
wejoncy: Done.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
YUNQIUGUO: ah just a general question wondering:

```
    private var TitleText: TextView? = null
    private var ArticleText: TextView? = null
    private var AnswerText: TextView? = null
    private var QuestionText: TextView? = null
    private var DoQAButton: Button? = null
```
is there anything in this list can be made non-nullable? as force operator `!!` for converting is recommended to generally avoid unless there's a must.

wejoncy: I see. Thanks. 'Lazy init' can make it.

ci_build/azure_pipelines/mobile-examples-pipeline.yml
```diff
@@ -280,3 +280,45 @@ jobs:
...
+  - bash: |
+      set -e
+      bash ./prepare_model.sh
+    workingDirectory: 'mobile/examples/question_answering/android'
+    displayName: "Generate model"
+    
+  - script: |
+      python3 ./ci_build/python/run_android_emulator.py \
+        --android-sdk-root ${ANDROID_SDK_ROOT} \
+        --create-avd --system-image "system-images;android-30;google_apis;x86_64" \
```
edgchen1, line 299: "default" image seems to be stabler for CI tests
```suggestion
        --create-avd --system-image "system-images;android-30;default;x86_64" \
```

ci_build/azure_pipelines/mobile-examples-pipeline.yml
```diff
@@ -280,3 +280,45 @@ jobs:
...
+  - task: JavaToolInstaller@0
+    displayName: Use jdk 11
+    inputs:
+      versionSpec: "11"
+      jdkArchitectureOption: "x64"
+      jdkSourceOption: "PreInstalled"
+
+  - bash: |
+      set -e
+      chmod +x ./gradlew
```
edgchen1, line 313: nit: can we just make the checked in gradlew file executable?
https://git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---chmod-x

wejoncy: Sure.Done.

mobile/examples/question_answering/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Thu Dec 29 10:51:23 PST 2022
+distributionBase=GRADLE_USER_HOME
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
```
edgchen1, line 3: could we also include the checksum?
https://docs.gradle.org/current/userguide/gradle_wrapper.html#configuring_checksum_verification
```suggestion
distributionSha256Sum=8cc27038d5dbd815759851ba53e70cf62e481b87494cc97cfd97982ada5ba634
distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
```

mobile/examples/question_answering/android/prepare_model.py
```diff
@@ -0,0 +1,64 @@
...
+    config = transformers.AutoConfig.from_pretrained(model_name)
+    model = transformers.MobileBertForQuestionAnswering.from_pretrained(model_name)
+    onnx_config = transformers.models.mobilebert.MobileBertOnnxConfig(config, "question-answering")
+
+    model_path = Path('app/src/main/res/raw/csarron_mobilebert_uncased_squad_v2.onnx')
+    onnx_inputs, onnx_outputs = transformers.onnx.export(tokenizer, model, onnx_config, 16, model_path)
+    return model_path
+
+
+def quantize_bert(model_path: Path):
```
YUNQIUGUO, line 26: nit: bert might be a bit ambiguous, would it make sense to just say `quantize_model`?

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+                    "without chewing it. After that they are not able to move," +
+                    " and they sleep through the six months that they need for digestion.\""
+        );
+        QuestionText.setHint("From which book did I see a magnificent picture?")
+
+        // Initialize Ort Session and register the onnxruntime extensions package that contains the custom operators.
+        // Note: These are used to decode the input image into the format the original model requires,
+        // and to encode the model output into png format
+        val sessionOptions: OrtSession.SessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
```
YUNQIUGUO, line 50: When the super resolution app is set up, yes we were using this approach to setup registerCustomOpLibrary, but for now maybe we can try to update to registering custom op using a specified function name i.e. `session_options.RegisterCustomOpsUsingFunction("RegisterCustomOps")`
(which is a suggested more unified way in all mobile platforms.)

sample usage:
https://github.com/microsoft/onnxruntime/blob/34175f0b7cd949b6b719633f68fb223fa129fee5/onnxruntime/test/custom_op_registration/test_registercustomops.cc#L45-L47

original pr for reference: https://github.com/microsoft/onnxruntime/pull/14177

hopefully it does not require any additional setup. we should update the SuperRes one as well.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+            "we are introduced to the narrator, a pilot, and his ideas about grown-ups." +
+                    "Once when I was six years old I saw a magnificent picture in a book," +
+                    " called True Stories from Nature, about the primeval forest. " +
+                    "It was a picture of a boa constrictor in the act of swallowing an animal. " +
+                    "Here is a copy of the drawing." +
+                    "In the book it said: \"Boa constrictors swallow their prey whole, " +
+                    "without chewing it. After that they are not able to move," +
+                    " and they sleep through the six months that they need for digestion.\""
+        );
+        QuestionText.setHint("From which book did I see a magnificent picture?")
```
YUNQIUGUO, line 44: not a comment, but assume `setHint` comes from Android `TextView`? wondering if iOS Swift has such equivalent features, seems like no.

that's okay should definitely be able to find out a work around in Swift. Good to see a hint will show when the input text box is null.

mobile/examples/question_answering/android/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```bash
+Model will be placed at `mobile\examples\question_answering\android\app\src\main\res\raw` automatically.
+```bash
+cd mobile/examples/question_answering/android
+bash prepare_models.sh
+```
+
+
+### Step 3: Connect Android Device and Run the app
```
YUNQIUGUO, line 49: nit: typo and format here.

YUNQIUGUO (commented):
Overall LGTM, just some minor comments. also thanks for working on adding the shared model python script.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
wejoncy: It only works with embeded ort-extensions, is that right?

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
wejoncy: Yeah, It comes from TextView.

It seems be PlaceHolder in IOS UITextView and It does't have the default method to use it though. I found some material which may work it through.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
YUNQIUGUO: I think the call of using function name should work with the current setup. but I can try with the super resolution android one and see if anything unexpected happens.

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
natke: How hard would it be to modify the app so that the user can enter a question and a context?

wejoncy: Not too hard. Do we want to support entering context in this example APP?

mobile/examples/question_answering/android/app/src/main/java/ai/onnxruntime/example/question_answering/MainActivity.kt
wejoncy: handle = handle ? handle : RTLD_DEFAULT; // onnxruntime/core/platform/posix/env.cc:561

So `dlsym` will try to find the symbol recursively by its depended library.

If we didn't build ort with ort-exts, it's somhow unlikey to find it, isn't it?

YUNQIUGUO (approved):

Issue #212, opened 2023-02-27
Difficulty compiling/running the `c_cxx` examples. MNIST example works on Visual Studio 2022


napulen:
Hello.

I had a few problems building the `c_cxx` examples provided in this repository. After some struggle, I got the `MNIST` example to work and can confirm that it runs well on Visual Studio 2022 with a CPU provider.

Here's a summary of the effort, following the instructions in the `README`.

My environment is:
- Win11 with Visual Studio 2022
- cmake  v3.26.0-rc4
- no libpng

I tried the Option 2 discussed in the `README`, building the `onnxruntime` from source. I could successfully build that using `Visual Studio 17 2022` as the CMAKE generator. The `.dll` and `.lib` are generated and installed under `C:\Program Files\onnxruntime`, as well as the `include/` headers. No problems to report.

When building the examples, however, things didn't go well. Specifically, trying the following within `onnxruntime-inference-examples/c_cxx/build`:

```batch
cmake .. -A x64 -T host=x64 -DONNXRUNTIME_ROOTDIR="C:\Program Files\onnxruntime"

msbuild onnxruntime_samples.sln /p:Configuration=Release
```

results in the following compilation error:

```
"D:\nnpolesl\onnx\onnxruntime-inference-examples\c_cxx\build\onnxruntime_samples.sln" (default target) (1) ->
"D:\nnpolesl\onnx\onnxruntime-inference-examples\c_cxx\build\imagenet\image_classifier.vcxproj.metaproj" (default targe
t) (7) ->
"D:\nnpolesl\onnx\onnxruntime-inference-examples\c_cxx\build\imagenet\image_classifier.vcxproj" (default target) (10) -
>
(ClCompile target) ->
  D:\nnpolesl\onnx\onnxruntime-inference-examples\c_cxx\include\providers.h(5,10): fatal  error C1083: Cannot open incl
ude file: 'cpu_provider_factory.h': No such file or directory [D:\nnpolesl\onnx\onnxruntime-inference-examples\c_cxx\bu
ild\imagenet\image_classifier.vcxproj]
```

It seems that the compiler is failing to find some headers. After closer inspection, I noticed that the `include/` folder is organized differently in the installation from source than in the redistributable `.zip` files located here: https://github.com/microsoft/onnxruntime/releases/tag/v1.14.0

In the latest redistributable release, the `include/` folder looks like this:
```
[onnxruntime-win-x86-1.14.0.zip]
include/
[...]

PR #215, opened 2023-03-13
Fix some issues with the C_CXX examples

skottmckay:
See #212 for more details.

- Cleanup the providers setup
  - remove unused providers
  - remove unnecessary include of the CPU EP factory header
  - Update cmake setup to clarify support for directory structure from pre-built package as well as a local install
- Copy ORT dlls to the output dir so an onnxruntime.dll in the system path doesn't break things at runtime
- Add mnist.onnx and copy to build dir
- Update min cmake version to match ORT
- Update to new cmake default install path of "C:/Program Files" as per https://cmake.org/cmake/help/v3.24/variable/CMAKE_INSTALL_PREFIX.html
- misc. cleanups to fix warning and line lengths


c_cxx/README.md
```diff
@@ -10,59 +10,69 @@ This directory contains a few C/C++ sample applications for demoing onnxruntime
 # How to build
 
 ## Prerequisites
-1. Visual Studio 2015/2017/2019
-2. cmake(version >=3.13)
+1. Visual Studio 2019 or 2022
+2. cmake (version >= 3.24)
```
edgchen1, line 14: should CMakeLists.txt be updated too?
https://github.com/microsoft/onnxruntime-inference-examples/blob/22fe8a46a2611602d28ea9d7fe8ced5fa5e37747/c_cxx/CMakeLists.txt#L4
there are some other CMakeLists.txt's specifying 3.13 too

c_cxx/CMakeLists.txt
```diff
@@ -27,16 +23,25 @@ set(CMAKE_CXX_STANDARD 17)
...
+    set(ONNXRUNTIME_ROOTDIR "C:/Program Files/onnxruntime")
+else()
+    set(ONNXRUNTIME_ROOTDIR "/usr/local")
   endif()
 endif()
 
+# The ORT package has a different include directory structure to a local install via cmake.
+# We added the path for the pre-built package above. Add the path for a local install to support either usage.
+# TODO: If we want to support additional EPs being loadable from a local install we also need to add EP specific
+# directories under /include/onnxruntime/core/providers
```
edgchen1, line 35: would this be needed if the onnxruntime_USE_CUDA or onnxruntime_USE_TENSORRT options are set?

c_cxx/MNIST/CMakeLists.txt
```diff
@@ -4,5 +4,9 @@
...
 
 add_executable(mnist MNIST.cpp)
+target_link_options(mnist PRIVATE "/SUBSYSTEM:WINDOWS")
 
-target_link_options(mnist PRIVATE "/SUBSYSTEM:WINDOWS")
\ No newline at end of file
+copy_ort_dlls(mnist)
+
+add_custom_command(TARGET mnist POST_BUILD
+                   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_SOURCE_DIR}/MNIST/mnist.onnx ${CMAKE_BINARY_DIR}/MNIST)
```
edgchen1, line 12: could we use TARGET_FILE_DIR here?
```suggestion
                   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_SOURCE_DIR}/MNIST/mnist.onnx $<TARGET_FILE_DIR:mnist>)
```

c_cxx/MNIST/readme.md
```diff
@@ -0,0 +1 @@
+See [ReadMe.md](../README.md) for build instructions.
```
edgchen1, line 1: does this need to be run from the same directory that the executable is in? e.g., to find mnist.onnx. might be worth calling that out if so.

c_cxx/README.md
```diff
@@ -10,59 +10,69 @@ This directory contains a few C/C++ sample applications for demoing onnxruntime
...
-You may append "-Donnxruntime_USE_CUDA=ON" or "-Donnxruntime_USE_DML=ON" to the last command args if your onnxruntime binary was built with CUDA or DirectML support respectively.
+You can open and build the solution using Visual Studio,
 
-You can then either open the solution in a Visual Studio and build it from there
 ```bat
 devenv onnxruntime_samples.sln
 ```
-Or build it using msbuild
+
+or build it from the command line using msbuild
```
edgchen1, line 74: using `cmake --build` could be more cross platform. but this readme seems pretty windows-specific now.

c_cxx/README.md
skottmckay: Went back to 3.13 as the CI doesn't have 3.24. Missed reverting this change.

c_cxx/README.md
skottmckay: Yeah. All the info here is Windows specific. At some point it would be good to fill in similar info for other platforms, but the goal of this PR is just to fix the existing setup that's broken.

c_cxx/MNIST/CMakeLists.txt
skottmckay: I set it up to copy to the default working directory VS uses when running tests - which is `$(ProjectDir)` in the VS values and is different to TARGET_FILE_DIR.

Updated to override the default working directory in VS to OutDir which is the same as TARGET_FILE_DIR.

c_cxx/CMakeLists.txt
```diff
@@ -27,16 +23,25 @@ set(CMAKE_CXX_STANDARD 17)
...
-    set(ONNXRUNTIME_ROOTDIR "C:/Program Files (x86)/onnxruntime")
-  else()
-    include_directories("/usr/local/include/onnxruntime")
+    set(ONNXRUNTIME_ROOTDIR "C:/Program Files/onnxruntime")
+else()
+    set(ONNXRUNTIME_ROOTDIR "/usr/local")
   endif()
 endif()
 
+# The ORT package has a different include directory structure to a local install via cmake.
```
edgchen1, line 32: > The ORT package has a different include directory structure to a local install via cmake.

is this discrepancy by design, or something to fix in ORT?

c_cxx/MNIST/CMakeLists.txt
```diff
@@ -4,5 +4,11 @@
...
 
 add_executable(mnist MNIST.cpp)
+target_link_options(mnist PRIVATE "/SUBSYSTEM:WINDOWS")
 
-target_link_options(mnist PRIVATE "/SUBSYSTEM:WINDOWS")
\ No newline at end of file
+copy_ort_dlls(mnist)
+
+# change working directory for VS debugger to same directory the binary is created in, and copy the model there.
+set_property(TARGET mnist PROPERTY VS_DEBUGGER_WORKING_DIRECTORY "$(OutDir)")
```
edgchen1, line 12: nice! good to know this is configurable in CMake

edgchen1 (approved):

c_cxx/README.md
edgchen1: got it. we could install a more recent CMake on the CI machines in an earlier pipeline step. on the other hand, a lower version requirement may be more accommodating to users.

skottmckay: yeah - I thought about the former and decided the latter was probably more important

c_cxx/CMakeLists.txt
skottmckay: I don't think we generally expect users to be building against ORT source directly. I'd lean towards the examples only using released packages for simplicity. 

It's also a side effect of the old way of enabling additional EPs (which also doesn't seem core to introductory examples) requiring a header to define an exported symbol and that header was in the internal EP directory. I think it would be better to convert any EP we want to use in these examples to use the generic registration function in the C API (e.g. XNNPACK, SNPE and a few other EPs are using that now). That way the user only needs things that come from the /include/onnxruntime path.

PR #221, opened 2023-03-29
add squeezenet sample for qnn ep cpu backend

jywu-msft:
add a simple squeezenet example for QNN EP (cpu backend)
built and tested with Visual Studio 2022 ARM64 and QNN 2.8 for Win/ARM.

### DIFF
--- a/c_cxx/QNN_EP/README.md
+++ b/c_cxx/QNN_EP/README.md
@@ -1,0 +1,5 @@
+## How to run the application
+(Windows11/ARM64) Run ```run_qnn_ep_sample.bat``` with path to onnxruntime home directory and current onnxruntime-inference-examples\c_cxx\QNN_EP directory
+```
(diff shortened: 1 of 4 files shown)


c_cxx/QNN_EP/README.md
```diff
@@ -0,0 +1,5 @@
+## How to run the application
```
HectorSVC, line 1: Should provide more instruction such as where to get the test model, where to place the test model, and also the qnn libs?
Should we enable the HTP test with QDQ model generation step?

jywu-msft: this example is strictly for cpu backend as a quick example of how to integrate with ORT. the readme gives the commandline for running the batchfile. 
the batchfile downloads the squeezenet model from onnx model zoo, builds the sample app and runs it.

jywu-msft: htp/qdq can be a different sample (once we've finalized our optimal quantization workflow)

PR #233, opened 2023-04-29
Add a VC++ static analysis Github workflow

snnn:
Compare to the ADO pipelines we already have, this one can directly publish scan results to Github website for view. It will be better integrated with Github UI. 

Example output:

image


jchen351:
We could ge a dynamic analysis tool, such as valgrind too.

jchen351 (approved):

PR #237, opened 2023-05-11
add whisper example for onnxruntime-web

guschmue:


shalvamist:
looks good - cool demo

shalvamist (approved):

fs-eire (approved):

lxfater:
Why doesn’t the Whisper example provide timestamps? I’m new to this field, so could you please explain it kindly? @guschmue

guschmue:
the model I'm using doesn't support that yet but people are working on it.
I'll add it once it is supported.

PR #251, opened 2023-05-25
Update ONNX Runtime Version

snnn:
1. Update ONNX Runtime Version
2. Remove Windows_Build_Nightly build job because it is not stable. The build job fetches artifacts from our nuget nightly build pipeline, but the pipeline now often run with a different configuration that only generates Azure EP packages. When it happens, this repo's pipeline will not be able to get the prebuilt binaries it needs. Therefore I will remove this build job for now, and try to find a better to get the binaries.


natke (approved):

natke:
I wonder if we can have dependabot update the ORT version?

snnn:
> I wonder if we can have dependabot update the ORT version?

No. C++ does not have a standard package manager. Therefore there is no standard way to query if a package has a newer version. If we choose to use NuGet to manage such binaries, it might work.

PR #252, opened 2023-05-25
Object detection: work with Unix line endings in classes.txt file.

janimo:
When checked out on Linux the classes.txt assets file only has \n as line endings so that is how it gets packaged into the app. This change makes the splitting of that file work with such line endings too otherwise there will be a single long classification label instead of 80.

### DIFF
--- a/mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt
+++ b/mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt
@@ -99,9 +99,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun readClasses(): List<String> {
-        val classes_id = R.raw.classes
-        val classes_bytes = resources.openRawResource(classes_id).readBytes()
-        return String(classes_bytes).split("\r\n".toRegex())
+        return resources.openRawResource(R.raw.classes).bufferedReader().readLines()
     }
 
     private fun readInputImage(): InputStream {


mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt
```diff
@@ -101,7 +101,7 @@ class MainActivity : AppCompatActivity() {
     private fun readClasses(): List<String> {
         val classes_id = R.raw.classes
         val classes_bytes = resources.openRawResource(classes_id).readBytes()
-        return String(classes_bytes).split("\r\n".toRegex())
+        return String(classes_bytes).split("\r?\n".toRegex())
```
snnn, line 104: Can we use a https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html here instead? 

Or like this:

https://stackoverflow.com/questions/37308957/how-do-you-read-a-text-file-from-a-resources-folder-in-java-1-7

edgchen1: something like this should work:
https://github.com/microsoft/onnxruntime-inference-examples/blob/6c6304169a75420b83547e3fedcfa706ea8b2ec4/mobile/examples/image_classification/android/app/src/main/java/ai/onnxruntime/example/imageclassifier/MainActivity.kt#L150

janimo:
@microsoft-github-policy-service agree

mobile/examples/object_detection/android/app/src/main/java/ai/onnxruntime/example/objectdetection/MainActivity.kt
janimo: @snnn @edgchen1 updated.

edgchen1 (approved):

PR #261, opened 2023-06-14
Add LLaMa quantization example

mengniwang95:
Add LLaMa quantization example, which is quantized by Intel® Neural Compressor(INC) with SmoothQuant algorithm.

**Note:**
During environment preparation, please execute `pip install -i https://test.pypi.org/simple/ neural-compressor==2.2.dev20230614` after `pip install -r requirements.txt` to use the latest code.

INC’s v2.2 official release targets 6/21, after that you can skip above step for test.


yufenglee (approved):

aliozts:
Does this quantization work for `llama-2`?

mengniwang95:
> Does this quantization work for `llama-2`?

Hi, this works for llama-2. And according to my test, 0.7 is the best sq alpha with pile-10k for calibration and lambada_openai for evaluation.

PR #296, opened 2023-09-08
Example speech recognition mobile app using Azure custom op on Android

YUNQIUGUO:
Example mobile app using Azure custom op on Android

Create a version that calls the OpenAI Whisper endpoint using the Azure custom op based on the existing Whisper Android application.

This demo can successfully call the OpenAI endpoint and return a json response.

The project contains a workable model created from this script: https://github.com/microsoft/onnxruntime-extensions/blob/main/test/data/azure/create_openai_whisper_transcriptions.py


skottmckay:
> TBD whether we should create a separate directory for this sample or we can place it under mobile/example/speech_recognition/azure_op_usage_demo.

May want to restructure as we have /mobile/examples/speech_recognition/android using Whisper but iOS using wav2vec2.

Given Whisper is fairly well known, maybe the apps using it could be in /mobile/examples/whisper/android/{local|azure}, which would leave the old setup under examples/speech_recogniztion of the ios and model directories?

And technically it is calling OpenAI not Azure but I expect we'll change that once Azure OpenAI releases their Whisper endpoint hence the directory name of 'azure'.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    init {
+        val env = OrtEnvironment.getEnvironment()
+        val sessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+
+        session = env.createSession(modelBytes, sessionOptions)
+
+        // NOTE!! Modify the following line and input your own OpenAI AUTH_TOKEN for successful API request calls.
+        // By default, the behavior of the app will fail to build if no correct authToken provided.
```
edgchen1, line 43: I think it's no longer true that the app fails to build.

also, it is probably worth warning that one should not actually hardcode an API token in the code.

mobile/examples/whisper/android/azure/readme.md
```diff
@@ -0,0 +1,39 @@
...
+
+Example App Scrennshot - A Successful result returned from OpenAI endpoint:
+
+<img width=25% src="images/Img_successful_endpoint_res.png" alt="App Screenshot" />
+
+## Set up
+
+### Prerequisites
+
+See the general prerequisites [here](../../../README.md#General-Prerequisites).
```
edgchen1, line 15: the relative link needs to be updated with the new directory structure. same comment in local/readme.md

mobile/examples/whisper/android/azure/readme.md
```diff
@@ -0,0 +1,39 @@
+# Android Speech Recognition Example with Azure Custom Op
+
+This example creates a version to show how to use ORT to do speech recognition using the [Whisper](https://github.com/openai/whisper) model and calls the OpenAI Whisper endpoint using the Azure custom op. You MUST update the code to specify a valid OpenAI auth token. See [here](#prepare-an-openai-auth-token) for details.
+
+The application lets the user make an audio recording, then recognizes the speech from that recording and displays a transcript.
+
+Example App Scrennshot - A Successful result returned from OpenAI endpoint:
```
edgchen1, line 7:
```suggestion
Example App Screenshot - A Successful result returned from OpenAI endpoint:
```

mobile/examples/whisper/android/azure/readme.md
```diff
@@ -0,0 +1,39 @@
+# Android Speech Recognition Example with Azure Custom Op
+
+This example creates a version to show how to use ORT to do speech recognition using the [Whisper](https://github.com/openai/whisper) model and calls the OpenAI Whisper endpoint using the Azure custom op. You MUST update the code to specify a valid OpenAI auth token. See [here](#prepare-an-openai-auth-token) for details.
```
edgchen1, line 3:
```suggestion
This example shows how to use ORT to do speech recognition using the [Whisper](https://github.com/openai/whisper) model and calls the OpenAI Whisper endpoint using the Azure custom op. You MUST update the code to specify a valid OpenAI auth token. See [here](#prepare-an-openai-auth-token) for details.
```

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            val responseText = json.decodeFromString<Response>(recognizedText)
+            return Result(responseText.text, elapsedTimeInMs)
+        } catch (e: Exception) {
+            println("Error: ${e.message}")
+        }
+
+
+        val errorMsg = json.decodeFromString<ErrorResponse>(recognizedText)
+        return Result(errorMsg.error.message, elapsedTimeInMs)
```
edgchen1, line 76: it seems odd to just return the error message as the result. might be better to signal that there was an error.

skottmckay: Really the custom op needs to detect an error so the inference fails. Otherwise you can't add downstream nodes in the model as they'll be receiving bogus input. 

Given this is an example app, I'm fine with outputting the json as-is for now but if it's simple to add an alert or something feel free to.

YUNQIUGUO: > Really the custom op needs to detect an error so the inference fails. Otherwise you can't add downstream nodes in the model as they'll be receiving bogus input.

Got it.

Agree that it doesn't make much sense to return an error message but with the app showing `successful recognition`. Will add an alert and mark it as Error in the app.

YUNQIUGUO: Updated. Current behavior is inference result is allowed to be successful with an error response. The app will check if it's an error response. If so, show as `Error` and display the message from return result of openai endpoint:
Screenshot 2023-09-15 at 11 10 46 AM

mobile/examples/azure_custom_op_demo_usage/android/app/src/main/java/ai/onnxruntime/example/speechrecognition/TensorUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+
+internal fun createInt8Tensor(env: OrtEnvironment, data: ByteArray, shape: LongArray, type: OnnxJavaType): OnnxTensor {
+    return OnnxTensor.createTensor(env, ByteBuffer.wrap(data), shape, type)
+}
+
+internal fun createStringTensor(env: OrtEnvironment, data: Array<String>, shape: LongArray): OnnxTensor {
+    return OnnxTensor.createTensor(env, data, shape)
+}
+
+internal fun createFloatTensor(
```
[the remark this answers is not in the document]
YUNQIUGUO: removed the unused method.

mobile/examples/azure_custom_op_demo_usage/android/app/src/main/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt
```diff
@@ -0,0 +1,85 @@
...
+    init {
+        val env = OrtEnvironment.getEnvironment()
+        val sessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+
+        session = env.createSession(modelBytes, sessionOptions)
+
+        // NOTE!! Modify the following line and input your own OpenAI AUTH_TOKEN for succesful API request calls.
+        // By default, the behavior of the app will fail to build if no correct authToken provided.
+        val authToken = "Set this to your auth token and add a closing double quote;
```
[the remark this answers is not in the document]
YUNQIUGUO: updated to make it compile and running in the CI now.

mobile/examples/whisper/android/azure/readme.md
YUNQIUGUO: thanks for the catch up. updated.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt
```diff
@@ -0,0 +1,86 @@
...
+        val env = OrtEnvironment.getEnvironment()
+        val sessionOptions = OrtSession.SessionOptions()
+        sessionOptions.registerCustomOpLibrary(OrtxPackage.getLibraryPath())
+
+        session = env.createSession(modelBytes, sessionOptions)
+
+        // NOTE!! Modify the following line and input your own OpenAI AUTH_TOKEN for successful API request calls.
+        // By default, it will return an error message (given no correct authToken provided.)
+        // DO NOT hardcode an auth token here. See `User Settings` section in your OpenAI account for more detailed information
+        // to find your API Key.
```
skottmckay, line 45: nit: how someone wants to manage their auth token is up to them so I wouldn't say 'DO NOT'. We could maybe suggest they store it in https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt
```diff
@@ -0,0 +1,86 @@
...
+        val authToken = "Set this to your auth token (OpenAI API Key);"
+        val authTokenInput = OnnxTensor.createTensor(env, arrayOf(authToken), tensorShape(1.toLong()))
+        baseInputs = mapOf(
+            "auth_token" to authTokenInput
+        )
+    }
+
+    data class Result(val text: String, val inferenceTimeInMs: Long, val type: Boolean)
```
skottmckay, line 53:
```suggestion
    data class Result(val text: String, val inferenceTimeInMs: Long, val successful: Boolean)
```

`type` doesn't really convey what that the boolean is false is there was an error.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/AudioTensorSource.kt
```diff
@@ -0,0 +1,103 @@
...
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.util.concurrent.atomic.AtomicBoolean
+
+class AudioTensorSource {
+    companion object {
+        private const val sampleRate = 16000
+        private const val maxAudioLengthInSeconds = 30
+
+        fun fromRawWavBytes(rawBytes: ByteArray): OnnxTensor {
```
edgchen1, line 20: is the model input expected to be a wav file or raw PCM data? from this method name it seems like it is a wav file, but the `fromRecording` method seems to be returning a tensor of PCM data.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/AudioTensorSource.kt
```diff
@@ -0,0 +1,103 @@
...
+        fun fromRecording(stopRecordingFlag: AtomicBoolean): OnnxTensor {
+            val recordingChunkLengthInSeconds = 1
+
+            val minBufferSize = maxOf(
+                AudioRecord.getMinBufferSize(
+                    sampleRate,
+                    AudioFormat.CHANNEL_IN_MONO,
+                    AudioFormat.ENCODING_PCM_8BIT,
+                ),
+                2 * recordingChunkLengthInSeconds * sampleRate * 1
```
edgchen1, line 43: nit: the `* 1` doesn't seem to add much meaning. maybe remove it or use a constant named something like `bytesPerSample`.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/AudioTensorSource.kt
```diff
@@ -0,0 +1,103 @@
...
+                        recordingChunkLengthInSeconds * sampleRate,
+                        audioData.size - audioDataOffset
+                    )
+
+                    val readResult = audioRecord.read(
+                        audioData, audioDataOffset, numBytesToRead,
+                        AudioRecord.READ_BLOCKING
+                    )
+
+                    Log.d(MainActivity.TAG, "AudioRecord.read(float[], ...) returned $readResult")
```
edgchen1, line 75:
```suggestion
                    Log.d(MainActivity.TAG, "AudioRecord.read(byte[], ...) returned $readResult")
```

mobile/examples/whisper/android/azure/app/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+    id 'org.jetbrains.kotlin.android'
+    id 'kotlinx-serialization'  // This plug-in is used for parsing json string
+}
+
+android {
+    namespace 'ai.onnxruntime.example.speechrecognition'
+    compileSdk 33
+
+    defaultConfig {
+        applicationId "ai.onnxruntime.example.speechrecognition"
```
edgchen1, line 12: can we use a different applicationId for each example? and possibly the namespace too as it matches applicationId by default.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/TensorUtils.kt
```diff
@@ -0,0 +1,5 @@
+package ai.onnxruntime.example.speechrecognition
+
+// TODO: Add more tensor helper methods here if needed.
```
edgchen1, line 3: nit: I think we don't need this comment

mobile/examples/whisper/android/azure/app/src/androidTest/java/ai/onnxruntime/example/speechrecognition/SpeechRecognitionInstrumentedTest.kt
edgchen1, line 1: regarding directory structure, I think the other examples are following a convention of `example_type/android` and `example_type/ios`. should we do the same here? e.g., `mobile/examples/whisper/azure/android`

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/AudioTensorSource.kt
skottmckay: I believe the model expects a wav as input as the `audio_format` attribute specifies that.

The OpenAI endpoint seems to be flexible though and is able to determine the audio type ignoring what you said it was. Possibly it uses that info as a hint only. Despite that it strangely requires you to provide a filename with an accepted input format as the extension.

It would be better to send in a wav as the Azure OAI endpoint may not be as forgiving and it's confusing for a new user to make sense of the input when it doesn't match what the model appears to require.

mobile/examples/whisper/azure/android/app/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+dependencies {
+    implementation 'androidx.core:core-ktx:1.10.0'
+    implementation 'androidx.appcompat:appcompat:1.6.1'
+    implementation 'com.google.android.material:material:1.8.0'
+    implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
+
+    implementation ("org.jetbrains.kotlinx:kotlinx-serialization-json:1.2.2")
+
+    implementation 'com.microsoft.onnxruntime:onnxruntime-android:latest.release'
+//    implementation 'com.microsoft.onnxruntime:onnxruntime-extensions-android:latest.release'
```
YUNQIUGUO, line 45: will update these since extensions release are coming out now.

mobile/examples/whisper/android/azure/app/src/main/java/ai/onnxruntime/example/speechrecognition/AudioTensorSource.kt
YUNQIUGUO: thanks. - got the `fromRecording` method updated to work with a wav format file too.

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
```diff
@@ -0,0 +1,202 @@
...
+            try {
+                inputStream = FileInputStream(file)
+                val fileLength = file.length().toInt()
+                val fileData = ByteArray(fileLength)
+
+                inputStream.read(fileData)
+                return fileData
+            } finally {
+                inputStream?.close()
+            }
```
edgchen1, line 43: consider [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) instead of try/finally
https://kotlinlang.org/docs/idioms.html#java-7-s-try-with-resources

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
```diff
@@ -0,0 +1,202 @@
...
+                val outputStream = FileOutputStream(outputFilePath)
+
+                // Write the WAV header
+                outputStream.write(header)
+
+                // Write the PCM audio data
+                outputStream.write(pcmData)
+
+                outputStream.close()
+            } catch (e: IOException) {
```
edgchen1, line 64: maybe just let this exception propagate out

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
```diff
@@ -0,0 +1,202 @@
...
+        private fun convertPcmToWav(
+            pcmData: ByteArray,
+            sampleRate: Int,
+            bitsPerSample: Int,
+            channels: Int,
+            outputFilePath: String
+        ) {
+            try {
+                val header = createWavHeader(pcmData.size, sampleRate, bitsPerSample, channels)
+                val outputStream = FileOutputStream(outputFilePath)
```
edgchen1, line 55: consider `use` here too so `outputStream` is automatically closed

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
```diff
@@ -0,0 +1,202 @@
...
+
+            return header
+        }
+
+        fun fromRawWavBytes(rawBytes: ByteArray): OnnxTensor {
+            val rawByteBuffer = ByteBuffer.wrap(rawBytes)
+            // TODO handle big-endian native order...
+            if (ByteOrder.nativeOrder() != ByteOrder.LITTLE_ENDIAN) {
+                throw NotImplementedError("Reading Wav data is only supported when native byte order is little-endian.")
+            }
```
edgchen1, line 116: do we need to enforce that the host byte order is little endian? we are only dealing with raw bytes and not wider types where byte order is significant.

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
```diff
@@ -0,0 +1,202 @@
...
+
+                val rawByteBuffer = ByteBuffer.wrap(rawWavBytes)
+                if (ByteOrder.nativeOrder() != ByteOrder.LITTLE_ENDIAN) {
+                    throw NotImplementedError("Reading Wav data is only supported when native byte order is little-endian.")
+                }
+                rawByteBuffer.order(ByteOrder.nativeOrder())
+
+                val env = OrtEnvironment.getEnvironment()
+                return OnnxTensor.createTensor(
+                    env, rawByteBuffer, tensorShape((audioData.size + 44).toLong()), OnnxJavaType.UINT8)
```
edgchen1, line 191: could we call fromRawWavBytes() here?

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/MainActivity.kt
```diff
@@ -0,0 +1,188 @@
...
+
+            workerThreadExecutor.submit {
+                try {
+                    val audioTensor = resources.openRawResource(R.raw.self_destruct_button).use {
+                        AudioTensorSource.fromRawWavBytes(it.readBytes())
+                    }
+                    val result = audioTensor.use { speechRecognizer.run(audioTensor) }
+                    // Note: There's case where we can successfully conduct an ORT inference session call, but get an error
+                    // response from OpenAI endpoint. Check here if the return result is an error message. If so, set Error status
+                    // for the app.
```
edgchen1, line 122: maybe this comment makes more sense in the implementation of `handleResult`?

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/MainActivity.kt
```diff
@@ -0,0 +1,188 @@
...
+    private fun setError(exception: Exception) {
+        Log.e(TAG, "Error: ${exception.localizedMessage}", exception)
+        runOnUiThread {
+            statusText.text = "Error"
+            resultText.text = exception.localizedMessage
+        }
+    }
+
+    private fun setError(result: SpeechRecognizer.Result) {
+        Log.i(TAG, "Error: ${result.text}")
```
edgchen1, line 50: nit: can we also log at error level to be consistent with the other `setError`?
```suggestion
        Log.e(TAG, "Error: ${result.text}")
```

mobile/examples/whisper/azure/android/readme.md
```diff
@@ -0,0 +1,39 @@
...
+Additionally, you will need to be able to record audio, either on an emulator or a device.
+
+This example was developed with Android Studio Giraffe | 2022.3.1 Patch 1.
+It is recommended to use that version or a newer one.
+
+### Generate the model
+
+This is optional. Currently a usable model file is checked in.
+
+Model was created with [create_openai_whisper_transcriptions.py](https://github.com/microsoft/onnxruntime-extensions/blob/main/test/data/azure/create_openai_whisper_transcriptions.py) script. If any changes are required they should be simple adjust. (e.g. change default audio format from wav to mp3).
```
edgchen1, line 26: > If any changes are required they should be simple adjust. (e.g. change default audio format from wav to mp3).

were any changes required to generate the model? if so, it may be good to list them.

mobile/examples/whisper/azure/android/readme.md
```diff
@@ -0,0 +1,39 @@
...
+
+Copy the model to `app/src/main/res/raw/openai_whisper_transcriptions.onnx`.
+
+### Prepare an OpenAI Auth Token
+
+It's required for the app to provide an OpenAI auth token. Please update the authToken variable in `SpeechRecognizer init` (src/android/java/ai/onnxruntime/example/speechrecognition/SpeechRecognizer.kt) with your token.
```
edgchen1, line 32:
```suggestion
It's required for the app to provide an OpenAI auth token. Please update the authToken variable in `SpeechRecognizer init` (app/src/main/java/ai/onnxruntime/example/whisperAzure/SpeechRecognizer.kt) with your token.
```

mobile/examples/whisper/azure/android/readme.md
YUNQIUGUO: for the sake of model used in this sample app, no we are just using the model as created from the script.

mobile/examples/whisper/azure/android/app/src/main/java/ai/onnxruntime/example/whisperAzure/AudioTensorSource.kt
YUNQIUGUO: removed!

mobile/README.md
```diff
@@ -64,7 +63,15 @@ The example application accomplishes the task of recovering a high resolution (H
...
 ### Question Answering
+
 The example app gives a demo of introducing question answering models with pre/post processing into mobile scenario. Currently supports on platform Android and iOS.
 
 - [Android Question answering](examples/question_answering/android)
 - [iOS Question answering](examples/question_answering/ios)
+
+### Whisper (Cloud/Local)
+
+This example shows how to use ORT to do speech recognition using the [Whisper](https://github.com/openai/whisper) model. One version (Cloud) calls the OpenAI Whisper endpoint using the Azure custom op. (MUST update the code to specify a valid OpenAI auth token. ). The other uses a local Whisper model.
```
skottmckay, line 74: nit: I wouldn't include the info about the auth token here as this file seems to mainly provide high level descriptions of the examples. The need to create the token is covered by the readme for the actual example.

mobile/examples/whisper/azure/android/readme.md
```diff
@@ -0,0 +1,39 @@
...
+
+<img width=25% src="images/Img_successful_endpoint_res.png" alt="App Screenshot" />
+
+## Set up
+
+### Prerequisites
+
+See the general prerequisites [here](../../../../README.md#General-Prerequisites).
+
+Additionally, you will need to be able to record audio, either on an emulator or a device.
```
skottmckay, line 17: nit: we include pre-recorded audio don't we and recording your own audio is optional?

mobile/examples/whisper/azure/android/readme.md
skottmckay: Can we instead say the model generated by the script is checked in as that's clearer than 'a usable model'

skottmckay, line 26:
```suggestion
Model was created with [create_openai_whisper_transcriptions.py](https://github.com/microsoft/onnxruntime-extensions/blob/main/test/data/azure/create_openai_whisper_transcriptions.py) script. If any changes are required they should only be simple adjustments (e.g. change default audio format from wav to mp3).
```

mobile/examples/whisper/azure/android/readme.md
YUNQIUGUO: yea, updated in both whisper samples.

mobile/README.md
YUNQIUGUO: removed.

mobile/examples/whisper/azure/android/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+<img width=25% src="images/Img_successful_endpoint_res.png" alt="App Screenshot" />
+
+## Set up
+
+### Prerequisites
+
+See the general prerequisites [here](../../../../README.md#General-Prerequisites).
+
+Optionally, you will need to be able to record audio, either on an emulator or a device. In the example, a pre-recorded audio is provided.
```
skottmckay, line 17:
```suggestion
A pre-recorded audio file is provided in the app to simplify usage, and you can also record audio if you choose. 
```

skottmckay (approved):

PR #346, opened 2023-12-12
add mistral example

ajindal1:
Add mistral benchmark example


ajindal1:
@hanbitmyths It seems the failure is not related to the changes I am making.

hanbitmyths (approved):
For urgency for conference, let's check in first, then refine later.

PR #347, opened 2023-12-12
update the model path to reflect Onnx model zoo changes

HectorSVC:
1. Update the model path to reflect the Onnx model zoo changes. Models were moved to archive folder
2. Remove codes not used


jywu-msft (approved):

jywu-msft:
just noticed there were pipeline failures earlier. should we be concerned about them?

HectorSVC:
> just noticed there were pipeline failures earlier. should we be concerned about them?

We haven't enabled any build pipeline here for QNN EP.

jywu-msft (approved):

PR #350, opened 2023-12-14
Update run_capi_application.sh: allow using a local model file

snnn:
Update run_capi_application.sh: allow using a local model file


HectorSVC (approved):
:shipit:

PR #380, opened 2024-02-07
use get_qnn_qdq_config to generate QDQ model for QNN EP

HectorSVC:
use get_qnn_qdq_config to generate QDQ model for QNN EP


jywu-msft (approved):

jywu-msft:
thanks for this!

PR #385, opened 2024-02-26
improve the segment-anything example

guschmue:


js/segment-anything/webpack.config.js
```diff
@@ -1,27 +1,52 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license
+import CopyWebpackPlugin from 'copy-webpack-plugin';
```
satyajandhyala, line 1: Is copyright removed intentionally?

fs-eire:
maybe not using any bundler is a better idea. this helps to remove those unnecessary files (webpack.conf.js, package.json) with upcoming security warnings.

js/segment-anything/webpack.config.js
guschmue: just a std webpack.config

guschmue:
Going to add support for slimsam soon which will need some extra processing at build time so very soon I'll need at least package.json.

fs-eire (approved):

PR #436, opened 2024-05-29
Fix wording in MainActivity.java

Leteno:
The fileNames mismatch with urls


edgchen1:
Thanks for the PR. This issue was fixed by #435.

PR #449, opened 2024-07-15
Updated Phi-3 Android App using Java API

fionabos:
Updated "Local Chatbot on Android with Phi-3, ONNX Runtime Mobile and ONNX Runtime Generate() API" project from C++ to java using the generate() Java API.

Main changes:
- updated model downloader and file downloads to download all packages at once instead of one at a time
- added scroll, auto-scroll, adjustable response bubble size to responses
- updated prompt template to give shorter answers
- disabled send function and added message while model is downloading to prevent crashing


mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -85,8 +108,29 @@ public void onClick(View v) {
                     @Override
                     public void run() {
                         try {
-                            genAIWrapper.run(promptQuestion_formatted);
-                        } catch (GenAIException e) {
+                            TokenizerStream stream = tokenizer.createStream();
```
edgchen1, line 111: the ort-genai classes that implement AutoCloseable should be `close()`d or the resources will leak.
https://stackoverflow.com/questions/56112598/proper-way-to-close-an-autocloseable

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -85,8 +108,29 @@ public void onClick(View v) {
...
+                 
+                                int token = generator.getLastTokenInSequence(0);
+                 
+                                tokenListener.accept(stream.decode(token));
+                            }
+
+                            //generator.close();
+                        }
+                        catch (GenAIException e) {
                             throw new RuntimeException(e);
```
edgchen1, line 134: having an uncaught exception from a thread may not behave well. maybe we should at least log it. even better if the error can be shown to the user.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -56,10 +67,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        //enable scrolling and resizing of text boxes
+        generatedTV.setMovementMethod(new ScrollingMovementMethod());
+        getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE);
+
         // adding on click listener for send message button.
         sendMsgIB.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
+                if (model == null) {
```
edgchen1, line 80: should we just disable the button until the model is ready?

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -160,17 +202,34 @@ private void downloadModels(Context context) throws GenAIException {
...
+                    private long pctDone = 0;
+                    @Override
+                    public void onDownloadProgress(long bytesDone, long bytesTotal) {
+                        if (bytesTotal > 0) {
+                            long newPctDone = bytesDone * 100 / bytesTotal;
+                            if (newPctDone > pctDone) {
+                                pctDone = newPctDone;
+                                Log.d(TAG, "Download" + fileName + ": " + pctDone
+                                    + "% of " + (bytesTotal/1024) + " KB");
+                            }
```
edgchen1, line 223: this seems like it could be simpler. I think ModelDownloader won't call this twice with the same bytesDone value so you may not need the newPctDone > pctDone check. if all we do is log, then maybe it is simpler to just do that in ModelDownloader.

mobile/examples/phi-3-vision/android/app/libs/onnxruntime-genai-0.4.0-dev.jar
fionabos, line 1: remove

fionabos: done

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: Button is enabled to give error message that model is still downloading to prevent app from crashing. I made a download progress meter on the Phi-3 Vision app that replaced this functionality which can be brought over but would need to update UX to include this.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -84,9 +108,44 @@ public void onClick(View v) {
...
+                            generator = new Generator(model, generatorParams);
+
+                            while (!generator.isDone()) {
+                                generator.computeLogits();
+                                generator.generateNextToken();
+                 
+                                int token = generator.getLastTokenInSequence(0);
+                 
+                                tokenListener.accept(stream.decode(token));
+                            }
```
skottmckay, line 135: Do we need to show all steps here, or would the simpler iteration pattern be better

```
            Generator generator = new Generator(model, generatorParams)) {
          // iterate (which calls computeLogits, generateNextToken, getLastTokenInSequence and
          // isDone)
          for (int token_id : generator) {
            // decode and call listener
            String token = stream.decode(token_id);
            listener.accept(token);
          }
```

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
edgchen1: `model` and `tokenizer` are possibly set by another thread, i.e., when the download complete callback initializes them.

we rely on both of them to be initialized here, but it is possible this check is hit (`if (model != null)`) between the assignment to `model` and the assignment to `tokenizer`.

disabling the button until they are ready is one simple way to avoid thread-safety issues. alternatively, maybe making those variables `volatile` and checking both `model` and `tokenizer` here would also work.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -84,9 +108,44 @@ public void onClick(View v) {
...
+
+                            generatorParams = model.createGeneratorParams();
+                            //examples for optional parameters to format AI response
+                            //generatorParams.setSearchOption("length_penalty", 1000);
+                            //generatorParams.setSearchOption("max_length", 500);
+
+                            encodedPrompt = tokenizer.encode(promptQuestion_formatted);
+                            generatorParams.setInput(encodedPrompt);
+
+                            generator = new Generator(model, generatorParams);
```
skottmckay, line 126: Can we do all this in the `try` so the close is automatically handled?

e.g. 
```java
try (stream = tokenizer.createStream();
     generatorParams = model.createGeneratorParams();
     ....) {
```

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -84,9 +108,44 @@ public void onClick(View v) {
...
+                            generatorParams.close();
+
+                        }
+                        catch (GenAIException e) {
+                            Log.e(TAG, "Exception occurred during model query: " + e.getMessage());
+                            if (generator != null) generator.close();
+                            if (encodedPrompt != null) encodedPrompt.close();
+                            if (stream != null) stream.close();
+                            if (generatorParams != null) generatorParams.close();
                             throw new RuntimeException(e);
```
skottmckay, line 149: Do we need to convert this to RuntimeException or could we just rethrow?

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -101,99 +160,98 @@ public void run() {
...
-        try {
-            genAIWrapper.close();
-        } catch (Exception e) {
-            Log.e(TAG, "exception from closing genAIWrapper", e);
-        }
-        genAIWrapper = null;
+        tokenizer.close();
+        tokenizer = null;
+        model.close();
+        model = null;
```
skottmckay, line 166: nit: is there any need to set to null if onDestory is the last point the instance is usable?

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/ModelDownloader.java
```diff
@@ -14,52 +17,88 @@
...
-            callback.onDownloadComplete();
+          long begin = System.currentTimeMillis();
+
+          byte[] buffer = new byte[4096];
+          int bytesRead;
+
+          while ((bytesRead = inputStream.read(buffer)) != -1) {
+            outputStream.write(buffer, 0, bytesRead);
+            if (callback != null) {
+              callback.onProgress(totalBytesRead, totalBytesRead + bytesRead, totalDownloadBytes);
```
edgchen1, line 68: I think we can assume that `bytesRead` will always be greater than zero within this loop. i.e., `onProgress` will be called only if there is actually progress.
https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-

mobile/examples/phi-3/android/app/build.gradle.kts
```diff
@@ -14,15 +14,10 @@ android {
...
-        externalNativeBuild {
-            cmake {
-                cppFlags += "-std=c++17"
-            }
-        }
 
         ndk {
             //noinspection ChromeOsAbiSupport
-            abiFilters += listOf("arm64-v8a", "x86_64")
+            abiFilters += listOf("arm64-v8a")
```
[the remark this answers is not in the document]
fionabos: Yes, x86_64 is needed for the emulator since almost all the available virtual device API's, including the default API 35 uses the x86_64 ABI. However, while this support is nice to include, the app rarely runs on the emulator usually crashing immediately. The few times I was able to get it to work, it produces responses very slowly (1 token every second or two). Don't know if there's anything that can be done on our end about that or not.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -56,10 +68,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        //enable scrolling and resizing of text boxes
+        generatedTV.setMovementMethod(new ScrollingMovementMethod());
+        getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE);
+
         // adding on click listener for send message button.
         sendMsgIB.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
+                if (model == null) {
+                    // if the edit text is empty display a toast message.
```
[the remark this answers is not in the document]
fionabos: updated!

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -101,99 +160,98 @@ public void run() {
...
+                "phi3-mini-4k-instruct-cpu-int4-rtn-block-32-acc-level-4.onnx",
+                "phi3-mini-4k-instruct-cpu-int4-rtn-block-32-acc-level-4.onnx.data",
+                "special_tokens_map.json",
+                "tokenizer.json",
+                "tokenizer.model",
+                "tokenizer_config.json");
+
+        List<Pair<String, String>> urlFilePairs = new ArrayList<>();
+        for (String file : files) {
+            if (/*file.endsWith(".data") ||*/ !fileExists(context, file)) {
```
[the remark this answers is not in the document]
fionabos: removed

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: I did this for readability since this is meant to be a tutorial for mobile developers.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: just followed existing syntax on previous example.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: not sure. did this to make sure for garbage collection.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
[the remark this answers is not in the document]
fionabos: removed.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/ModelDownloader.java
fionabos: top comment: model != null should be an atomic comparison, so I think it should be safe in this context. The only dangerous thing is if it becomes null after being non null, but those should be all processed in the UI thread.

bottom comment: I think that's the correct assumption. It becomes -1 when we reach the end.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -84,9 +108,44 @@ public void onClick(View v) {
...
+
+                            generator.close();
+                            encodedPrompt.close();
+                            stream.close();
+                            generatorParams.close();
+
+                        }
+                        catch (GenAIException e) {
+                            Log.e(TAG, "Exception occurred during model query: " + e.getMessage());
+                            if (generator != null) generator.close();
```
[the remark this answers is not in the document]
fionabos: moved close to finally block, this allows us to rethrow the exception.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: I thought close would be automatically handled anyway since it's AutoClosable, based on what I read online. But if you think it's better to do it this way, please make the change.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
fionabos: You are right, I'll replace with tokenizer == null comparison to make sure we don't hit a race condition. This comparison should be atomic so I don't think we need to move allocator to UI thread.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
edgchen1: I believe `close()` is only automatically called when using a try-with-resources block (that construct shown in Scott's comment).
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/AutoCloseable.html

edgchen1 (approved):

PR #454, opened 2024-08-09
Fix mobile examples CI pipeline

edgchen1:
Fix the mobile examples CI pipeline.

In particular:
- Move Android examples to run on Linux build agents
- Copy over Android emulator helper scripts from ORT at commit 5e66fcc703
- Fix model download script for iOS object detection example
- Update Podfiles to use iOS 13 as that is the new minimum version required by ORT
- Refactor CI pipeline

### DIFF
--- a/ci_build/azure_pipelines/mobile-examples-pipeline.yml
+++ b/ci_build/azure_pipelines/mobile-examples-pipeline.yml
@@ -46,8 +46,6 @@ stages:
     pool:
       vmImage: "macOS-12"
 
-    dependsOn: ProcessParameters
-
     variables:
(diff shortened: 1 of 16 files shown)


ci_build/azure_pipelines/mobile-examples-pipeline.yml
edgchen1, line 1: I suggest hiding whitespace diffs when viewing this file

ci_build/python/run_android_emulator.py
edgchen1, line 1: changes under `ci_build/python` are mostly copied from corresponding ORT files at `tools/python`

skottmckay (approved):

PR #459, opened 2024-08-21
Adding examples for react-native-phi3

Vibrat:
Hi, 

I've been creating [react-native-phi3](https://www.npmjs.com/package/react-native-phi3) which wraps `onnxruntime-genai` for `phi3` inference. I think it would be great to have an example of it here, so this PR is created for it. 

Would be nice if someone can take a look on this.


Vibrat:
@microsoft-github-policy-service agree [company="{your company}"]

PR #467, opened 2024-10-03
Vraspar/phi 3 ios update

vraspar:
Update Phi-3 iOS application:
 
- Modify ReadME and add sample screenshot
- Add logging and display token generation stats
- Use User input instead of fixed prompt
- UI Improvements


mobile/examples/phi-3/ios/LocalLLM/LocalLLM/ContentView.swift
```diff
@@ -1,27 +1,124 @@
...
+                let combinedText = tokens.joined(separator: "")
+                messages[lastIndex].text = combinedText
+            }
+        }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("TokenGenerationStats"))) { notification in
+            if let userInfo = notification.userInfo,
+               let totalTime = userInfo["totalTime"] as? Int,
+               let firstTokenTime = userInfo["firstTokenTime"] as? Int,
+               let tokenCount = userInfo["tokenCount"] as? Int {
+                stats = "Generated \(tokenCount) tokens in \(totalTime) ms. First token in \(firstTokenTime) ms."
```
edgchen1, line 94: could we also include something like this:

`token generation rate (tokens/second) = (tokenCount - 1) * 1000 / (totalTimeInMs - firstTokenTimeInMs)`

the token generation and prompt processing rates may be different, and it might be useful to get a sense of both.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
-
-  auto model = OgaModel::Create(modelPath);
-  auto tokenizer = OgaTokenizer::Create(*model);
-
-  NSString* promptString = [NSString stringWithFormat:@"<|user|>\n%@<|end|>\n<|assistant|>", input_user_question];
-  const char* prompt = [promptString UTF8String];
-
-  auto sequences = OgaSequences::Create();
-  tokenizer->Encode(prompt, *sequences);
+typedef std::chrono::high_resolution_clock Clock;
```
edgchen1, line 12: nit: consider steady_clock

https://en.cppreference.com/w/cpp/chrono/steady_clock
"This clock is not related to wall clock time (for example, it can be time since last reboot), and is most suitable for measuring intervals."

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
++ (void)generate:(nonnull NSString*)input_user_question {
+    NSLog(@"Starting token generation...");
+    
+    NSString* llmPath = [[NSBundle mainBundle] resourcePath];
+    const char* modelPath = llmPath.cString;
+    
+    // Log model creation
+    NSLog(@"Creating model ...");
+    auto model = OgaModel::Create(modelPath);
+    if (!model) {
```
edgchen1, line 24: the C++ API throws exceptions.

https://github.com/microsoft/onnxruntime-genai/blob/bcf55a6dc563bc8b356128b47504d59a21c5ef2f/src/ort_genai.h#L54-L60

https://github.com/microsoft/onnxruntime-genai/blob/bcf55a6dc563bc8b356128b47504d59a21c5ef2f/src/ort_genai.h#L65

might be simplest to just put most of this method into a `try {}`. propagating the error back to the UI would be a nice touch.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
-  auto generator = OgaGenerator::Create(*model, *params);
++ (void)generate:(nonnull NSString*)input_user_question {
+    NSLog(@"Starting token generation...");
+    
+    NSString* llmPath = [[NSBundle mainBundle] resourcePath];
+    const char* modelPath = llmPath.cString;
+    
+    // Log model creation
+    NSLog(@"Creating model ...");
+    auto model = OgaModel::Create(modelPath);
```
edgchen1, line 23: performance-wise, it's probably nicer to not re-create the model every time.

edgchen1 (commented):
nice improvements!

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/Simulator Screenshot - iPhone 16.png
skottmckay, line 1: ❤️

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
skottmckay: it should definitely be nicer as the first request is usually the slowest (and typically when doing perf testing we do a warmup query first and exclude that from timing data).

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
+        NSLog(@"Failed to create model.");
+        return;
+    }
+    
+    NSLog(@"Creating tokenizer...");
+    auto tokenizer = OgaTokenizer::Create(*model);
+    if (!tokenizer) {
+        NSLog(@"Failed to create tokenizer.");
+        return;
+    }
```
skottmckay, line 34: The tokenizer could also be created once and re-used as I believe it's tied to the model not the prompt so can be re-used.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
+    int tokenCount = 0;
+    
+    NSLog(@"Starting token generation loop...");
+    while (!generator->IsDone()) {
+        generator->ComputeLogits();
+        generator->GenerateNextToken();
+        
+        if (isFirstToken) {
+            NSLog(@"First token generated.");
+            firstTokenTime = Clock::now();
```
skottmckay, line 67: nit: logging takes time so we want to record firstTokenTime prior to logging. 

might also be interesting to include per-token timing in the log as that would help get a picture of performance throughout the generation phase. e.g. is the time per token consistent? if not is the variability random or does it gradually increase/decrease? that can provide hints as to potential causes of performance issues (if there are any).

if we do that we might want to accumulate per-token times in a list and log at the end, as inserting log calls for every token inside the loop could significantly affect overall time taken (esp. is the log call is synchronous which apparently NSLog is).

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -5,45 +5,109 @@
...
+        // Decode the new token
+        const char* decode_tokens = tokenizer_stream->Decode(seq[seq_len - 1]);
+        
+        // Check for decoding failure
+        if (!decode_tokens) {
+            NSLog(@"Token decoding failed.");
+            break;
+        }
+        
+        NSLog(@"Decoded token: %s", decode_tokens);
```
skottmckay, line 84: nit: might want a setting to control whether we log inside the loop (if we're after the best perf numbers possible we probably don't want to do that), or a way to exclude time in calls to NSLog from the total.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
vraspar: saving timings in array and logging them after loop is over

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -1,49 +1,143 @@
...
-  }
+        TimePoint endTime = Clock::now();
+        // Log token times
+        NSLog(@"Per-token generation times: %@", [self formatTokenTimes:tokenTimes]);
+
+        // Calculate metrics
+        auto totalDuration = std::chrono::duration_cast<std::chrono::milliseconds>(endTime - startTime).count();
+        auto firstTokenDuration = std::chrono::duration_cast<std::chrono::milliseconds>(firstTokenTime - startTime).count();
+
+        double promtProcTime =  (double)promptTokensCount / firstTokenDuration;
```
edgchen1, line 102: should this be `promptTokens * 1000.0 / firstTokenDuration`?

and `promptProcRate` may be a better name.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -1,49 +1,143 @@
...
+        double tokenGenRate = (double)(tokenTimes.size() - 1) * 1000.0 / (totalDuration - firstTokenDuration);
+
+        NSLog(@"Token generation completed. Total time: %lld ms, First token time: %lld ms, Total tokens: %zu", totalDuration, firstTokenDuration, tokenTimes.size());
+        NSLog(@"Prompt tokens: %zu, Prompt Processing Time: %f tokens/s", promptTokensCount, promtProcTime);
```
edgchen1, line 106:
```suggestion
        NSLog(@"Prompt tokens: %zu, Prompt Processing Rate: %f tokens/s", promptTokensCount, promtProcTime);
```

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -1,49 +1,143 @@
...
-
+#include <chrono>
+#include <vector>
 
 @implementation GenAIGenerator
 
+typedef std::chrono::steady_clock Clock;
+typedef std::chrono::time_point<Clock> TimePoint;
+static std::unique_ptr<OgaModel> model = nullptr;
+static std::unique_ptr<OgaTokenizer> tokenizer = nullptr;
```
edgchen1, line 15: nit: can we avoid using global variables? maybe add some instance variables to GenAIGenerator instead.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -1,143 +1,146 @@
...
+        isFirstToken = false;
+      }
+
+      // Get the sequence data and decode the token
+      const int32_t* seq = generator->GetSequenceData(0);
+      size_t seq_len = generator->GetSequenceCount(0);
+      const char* decode_tokens = tokenizer_stream->Decode(seq[seq_len - 1]);
+
+      if (!decode_tokens) {
+        @throw [NSException exceptionWithName:@"TokenDecodeError" reason:@"Token decoding failed." userInfo:nil];
```
edgchen1, line 89: throw a C++ exception within the C++ `catch`

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/GenAIGenerator.mm
```diff
@@ -1,143 +1,146 @@
...
+      }
+
+      // Measure token generation time excluding logging
+      TimePoint tokenEndTime = Clock::now();
+      auto tokenDuration = std::chrono::duration_cast<std::chrono::milliseconds>(tokenEndTime - tokenStartTime).count();
+      tokenTimes.push_back(tokenDuration);
```
edgchen1, line 95: nit: consider calling `tokenTimes.reserve()` first to avoid `std::vector` reallocations during the measurement.

mobile/examples/phi-3/ios/LocalLLM/LocalLLM/ContentView.swift
```diff
@@ -3,25 +3,143 @@
...
+    var text: String
+    let isUser: Bool
+}
+
 struct ContentView: View {
-    @ObservedObject var tokenUpdater = SharedTokenUpdater.shared
+    @State private var userInput: String = ""
+    @State private var messages: [Message] = []  // Store chat messages locally
+    @State private var isGenerating: Bool = false  // Track token generation state
+    @State private var stats: String = ""  // token genetation stats
```
edgchen1, line 17:
```suggestion
    @State private var stats: String = ""  // token generation stats
```

edgchen1 (approved):

PR #482, opened 2024-10-14
Remove old iOS example.

edgchen1:
Other iOS examples are located in mobile/examples. This one is old, has limited documentation, and requires manual setup to get working.

It may be confusing to have iOS examples in multiple places.


vraspar:
should we also migrate/remove example under swift directory ?

edgchen1:
> should we also migrate/remove example under swift directory ?

yeah, I was considering that too. I think so. we can probably replace it with a Swift example that uses the onnxruntime-objc Pod. I think we can do that in another PR.

vraspar (approved):

skottmckay (approved):

PR #515, opened 2025-05-09
iOS model tester app

edgchen1:
Add iOS model tester app for measuring performance of a model.

There is also some cross-platform native code in `mobile/examples/model_tester/common` that can hopefully be reused in other contexts, such as an Android app.

### DIFF
--- a/mobile/examples/model_tester/common/model_runner.cpp
+++ b/mobile/examples/model_tester/common/model_runner.cpp
@@ -211,19 +211,16 @@ RunResult Run(const RunConfig& run_config) {
   return run_result;
 }
 
-std::string GetRunSummary(const RunConfig& run_config, const RunResult& run_result) {
+std::string GetRunSummary(const RunConfig& /*run_config*/, const RunResult& run_result) {
   auto to_display_duration = []<typename Rep, typename Period>(std::chrono::duration<Rep, Period> d) {
     using DisplayPeriod = std::chrono::microseconds::period;
     using DisplayDuration = std::chrono::duration<Rep, DisplayPeriod>;
     return std::chrono::duration_cast<DisplayDuration>(d);
   };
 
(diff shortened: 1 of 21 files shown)


mobile/examples/model_tester/ios/ModelTester/readme.md
```diff
@@ -0,0 +1,11 @@
+This is a basic app that runs an arbitrary ONNX model. It can measure the model's performance.
+
+# Set Up Instructions
+
+Copy an ONNX model to `model/model.onnx`.
```
vraspar, line 5: Should we add support for .onnx_data files? Maybe out of scope for this PR

mobile/examples/model_tester/common/model_runner.cpp
```diff
@@ -0,0 +1,245 @@
...
+  auto input_values = GetModelInputValues(session);
+
+  auto output_names = GetModelOutputNames(session);
+  auto output_name_cstrs = GetCstrs(output_names);
+
+  auto run_options = Ort::RunOptions{};
+
+  // warmup
+  for (size_t i = 0; i < run_config.num_warmup_iterations; ++i) {
+    auto outputs = session.Run(run_options,
```
vraspar, line 196: should we also add option for EP to RunConfig?

edgchen1: good idea. how about starting with supporting the generic AppendExecutionProvider() interface? e.g., string provider name and provider option values.

vraspar: yeah, that sounds good!

mobile/examples/model_tester/ios/ModelTester/model/download_model.py
vraspar, line 1: I am writing android version of the app, and it seems that it would be nice to have the downloading model in common directory as well, so android application can share the model creation steps

Issue #511, opened 2025-03-03
deepseek android example


suyash-narain:
Hi,

is there an android example in the works to deploy deepseek model on onnxruntime android?
deepseek is supported on onnxruntime-genai v0.6.0 whereas the android example using phi3 is using v0.4.0-dev.

There have been considerable changes between these two versions with a lot of code for phi3 android example becoming deprecated if i use it with deepseek. 

Is there an android example using the latest version of onnxruntime-genai v0.6.0?

apparently, the Model class in onnxruntime-genai no longer have createTokenizer, createGeneratorParams methods?

thanks

vraspar:
I did create PR to update the ios app with new API while ago: https://github.com/microsoft/onnxruntime-inference-examples/pull/505

I will create PR to update the android example :)

PR #520, opened 2025-06-03
Update Phi3 Android App

vraspar:
Update the ONNX Runtime GenAI to `v0.8.1` and refactor MainActivity to use SimpleGenAI API. 

Fixes: https://github.com/microsoft/onnxruntime-inference-examples/issues/511


mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -131,77 +141,57 @@ public void onClick(View v) {
...
-
-                                if (numTokens == 0) { //first token
-                                    firstTokenTime = System.currentTimeMillis();
+                            AtomicLong firstTokenTime = new AtomicLong(startTime);
+                            AtomicInteger numTokens = new AtomicInteger(0);
+                            
+                            // Token listener for streaming tokens
+                            Consumer<String> tokenListener = token -> {
+                                if (numTokens.get() == 0) { // first token
+                                    firstTokenTime.set(System.currentTimeMillis());
```
edgchen1, line 159: reading and then writing two atomic variables is not thread-safe, even though the variables themselves are atomic. there's a [compareAndSet](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicLong.html#compareAndSet-long-long-) that might be better suited for this. e.g.:

```suggestion
                                firstTokenTime.compareAndSet(startTime, System.currentTimeMillis());
```

alternatively, if we can get some guarantee from the API that the listener will only be called from a single thread, as it is [now](https://github.com/microsoft/onnxruntime-genai/blob/b4740f1689058567f814c697e92cfc6e0895142b/src/java/src/main/java/ai/onnxruntime/genai/SimpleGenAI.java#L79-L83), we could avoid any thread-safety precautions entirely.

mobile/examples/phi-3/android/app/build.gradle.kts
```diff
@@ -52,6 +52,6 @@ dependencies {
 
     // ONNX Runtime with GenAI
     implementation("com.microsoft.onnxruntime:onnxruntime-android:latest.release")
-    implementation(files("libs/onnxruntime-genai-android-0.4.0-dev.aar"))
+    implementation(files("libs/onnxruntime-genai-android-0.8.1.aar"))
```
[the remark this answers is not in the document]
edgchen1: wonder if there's a way to have gradle download from github? it'd be nice to not have to keep a copy of the .aar in the repo. we don't have to figure that out in this PR though.

mobile/examples/phi-3/android/app/src/main/java/ai/onnxruntime/genai/demo/MainActivity.java
```diff
@@ -131,77 +141,56 @@ public void onClick(View v) {
...
-                            while (!generator.isDone()) {
-                                generator.computeLogits();
-                                generator.generateNextToken();
-                 
-                                int token = generator.getLastTokenInSequence(0);
-
-                                if (numTokens == 0) { //first token
-                                    firstTokenTime = System.currentTimeMillis();
+                            AtomicLong firstTokenTime = new AtomicLong(startTime);
+                            AtomicInteger numTokens = new AtomicInteger(0);
```
edgchen1, line 154: I believe these don't need to be atomic. now, the SimpleGenAI documentation says that the listener will not be called concurrently.
https://github.com/microsoft/onnxruntime-genai/blob/0a34c76e024cff85f0018f495e67de0bda417ffe/src/java/src/main/java/ai/onnxruntime/genai/SimpleGenAI.java#L61-L64

edgchen1 (approved):

PR #529, opened 2025-09-18
ORT tutorials and samples

gedoensmax:
# Description

This PR adds c_cxx samples showcasing the new 1.23.0 features in a tutorial format. I was tempted to make this the main c_cxx directory and move all the older samples to a subfolder. Please let me know what you think.

The idea was to make each of the samples self-contained at the cost of some minimal duplication. The samples should be buildable individually or jointly from the c_cxx root folder.
The numbering of the sample is a multiple of 10 to subsequently add smaller additions before or after a certain sample without changing the numbering of the existing samples. Higher numbers indicate more advanced concepts.

At the moment, binaries to ONNX Runtime libraries and execution providers must be specified at build time using CMake variables. Ideally, we can offer a way to automatically download the binaries once 1.23 is released.

The python API samples for device tensors were also refactored to use hardware independent tensors and a compile sample for EP context files was added.


snnn (approved):
Thanks. I will add them to CI later.

PR #536, opened 2025-09-29
[Plugin EP] Add python example for how to run a plugin EP

chilo-ms:
We have a [documentation ](https://onnxruntime.ai/docs/execution-providers/plugin-ep-libraries.html#using-a-plugin-ep-library)for showing how to use C++ API to run a plugin EP.
Add an example for python usage.

### DIFF
--- a/python/plugin_EP/README.md
+++ b/python/plugin_EP/README.md
@@ -1,0 +1,4 @@
+# Running Inference with a Plugin EP
+## Prerequisites
+- A dynamic/shared EP library that exports the functions `CreateEpFactories()` and `ReleaseEpFactory()`.
+- ONNX Runtime built as a shared library (e.g., `onnxruntime.dll` on Windows or `libonnxruntime.so` on Linux), since the EP library relies on the public ORT C API (which is ABI-stable) to interact with ONNX Runtime. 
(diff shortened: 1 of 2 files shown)


c_cxx/plugin_EP/README.md
```diff
@@ -0,0 +1,69 @@
+# Running Inference with a Plugin EP using C++ API
+## Prerequisites
+- ONNX Runtime version >= 1.23.0
+- A dynamic/shared EP library that exports the functions `CreateEpFactories()` and `ReleaseEpFactory()`.
+- ONNX Runtime built as a shared library (e.g., `onnxruntime.dll` on Windows or `libonnxruntime.so` on Linux), since the EP library relies on the public ORT C API (which is ABI-stable) to interact with ONNX Runtime.
+- The `onnxruntime_providers_shared.dll` (Windows) or `libonnxruntime_providers_shared.so` (Linux) is also required. When a plugin EP is registered, ONNX Runtime internally calls `LoadPluginOrProviderBridge`, which depends on this shared library to determine whether the EP DLL is a plugin or a provider-bridge.
+- If you are using a pre-built ONNX Runtime package, all required libraries (e.g., `onnxruntime.dll`, `onnxruntime_providers_shared.dll`, etc.) are already included.
```
adrianlizarraga, line 7: The `onnxruntime_providers_shared.dll` is not required for new EPs that are only using the new binary-stable APIs. If this example is only meant for new EPs, perhaps we don't need to mention it?

